PR #8, opened 2021-05-28
heal: Add MRF metrics

vadmeste:
Expose MRF information in heal background status.

### DIFF
--- a/heal-commands.go
+++ b/heal-commands.go
@@ -296,6 +296,17 @@ func (adm *AdminClient) Heal(ctx context.Context, bucket, prefix string,
 	return healStart, healTaskStatus, nil
 }
 
+// MRFStatus exposes MRF metrics of a server
+type MRFStatus struct {
+	BytesHealed uint64 `json:"bytes_healed"`
+	ItemsHealed uint64 `json:"items_healed"`
+
+	TotalItems uint64 `json:"total_items"`
+	TotalBytes uint64 `json:"total_bytes"`
+
+	Started time.Time `json:"started"`
+}
+
 // BgHealState represents the status of the background heal
 type BgHealState struct {
 	ScannedItemsCount int64
@@ -304,6 +315,9 @@ type BgHealState struct {
 
 	// SetStatus contains information for each set.
 	Sets []SetStatus `json:"sets"`
+
+	// Endpoint -> MRF Status
+	MRF map[string]MRFStatus `json:"mrf"`
 }
 
 // SetStatus contains information about the heal status of a set.
@@ -313,6 +327,7 @@ type SetStatus struct {
 	SetIndex     int    `json:"set_index"`
 	HealStatus   string `json:"heal_status"`
 	HealPriority string `json:"heal_priority"`
+	TotalObjects int    `json:"total_objects"`
 	Disks        []Disk `json:"disks"`
 }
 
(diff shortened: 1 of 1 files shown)


heal-commands.go
```diff
@@ -296,6 +296,17 @@ func (adm *AdminClient) Heal(ctx context.Context, bucket, prefix string,
...
 
+// MRFStatus exposes MRF metrics of a server
+type MRFStatus struct {
+	BytesHealed uint64 `json:"bytesHealed"`
+	ItemsHealed uint64 `json:"itemsHealed"`
+
+	TotalItems uint64 `json:"totalItems"`
+	TotalBytes uint64 `json:"totalBytes"`
+
+	Started time.Time `json:"started"`
```
harshavardhana, line 307: what does this signify?  - isn't it always Started?

heal-commands.go
```diff
@@ -321,18 +336,22 @@ type HealingDisk struct {
...
-	PoolIndex     int       `json:"pool_index"`
-	SetIndex      int       `json:"set_index"`
-	DiskIndex     int       `json:"disk_index"`
-	Endpoint      string    `json:"endpoint"`
-	Path          string    `json:"path"`
-	Started       time.Time `json:"started"`
-	LastUpdate    time.Time `json:"last_update"`
-	ObjectsHealed uint64    `json:"objects_healed"`
-	ObjectsFailed uint64    `json:"objects_failed"`
-	BytesDone     uint64    `json:"bytes_done"`
```
harshavardhana, line 334: Is this a breaking change? you are removing two fields? and its renamed as `Items` ? why ?

heal-commands.go
```diff
@@ -321,18 +336,22 @@ type HealingDisk struct {
...
+	Endpoint   string    `json:"endpoint"`
+	Path       string    `json:"path"`
+	Started    time.Time `json:"started"`
+	LastUpdate time.Time `json:"last_update"`
+
+	ObjectsTotalCount uint64 `json:"objects_total_count"`
+	ObjectsTotalSize  uint64 `json:"objects_total_size"`
+
+	ItemsHealed uint64 `json:"items_healed"`
+	ItemsFailed uint64 `json:"items_failed"`
```
harshavardhana, line 352: We an keep items as Objects?

heal-commands.go
```diff
@@ -296,6 +296,17 @@ func (adm *AdminClient) Heal(ctx context.Context, bucket, prefix string,
 	return healStart, healTaskStatus, nil
 }
 
+// MRFStatus exposes MRF metrics of a server
+type MRFStatus struct {
+	BytesHealed uint64 `json:"bytesHealed"`
+	ItemsHealed uint64 `json:"itemsHealed"`
+
+	TotalItems uint64 `json:"totalItems"`
```
harshavardhana, line 304: TotalObjects

heal-commands.go
```diff
@@ -296,6 +296,17 @@ func (adm *AdminClient) Heal(ctx context.Context, bucket, prefix string,
 	return healStart, healTaskStatus, nil
 }
 
+// MRFStatus exposes MRF metrics of a server
+type MRFStatus struct {
+	BytesHealed uint64 `json:"bytesHealed"`
+	ItemsHealed uint64 `json:"itemsHealed"`
```
harshavardhana, line 302:
```suggestion
	ObjectsHealed uint64 `json:"itemsHealed"`
```

harshavardhana: Also let's follow either `key_value` style or `keyValue` style - let's not mix both for JSON tags.

harshavardhana (changes requested):

heal-commands.go
vadmeste: > Is this a breaking change? you are removing two fields? and its renamed as `Items` ? why ?

emm yes, this is a breaking change. I just wanted to differentiate between objects and items.

Items means objects inside buckets and inside .minio.sys
Objects means only objects inside buckets.

Otherwise the user can be confused by seeing 100 objects healing while he has only 90 objects.

heal-commands.go
vadmeste: > what does this signify? - isn't it always Started?

Well, we need to show users that there is no MRF activity at a given time. Basically when the MRF queue is empty, then Started will be time sentinel.

So yes, the server will be able to show that there is no MRF activity (basically it means MRF queue for objects is empty)

heal-commands.go
vadmeste: I should keeping the old values and mark them as deprecated

vadmeste: > I should keeping the old values and mark them as deprecated

I kept those:
```
ObjectsHealed uint64 `json:"objects_healed"` // Deprecated July 2021
ObjectsFailed uint64 `json:"objects_failed"` // Deprecated July 2021
```

Though I don't think that anyone is using it right now. This only breaks people consuming health info in json format

harshavardhana: Yes lets keep it this way as subnet health analyzer might break.

harshavardhana (approved):

PR #12, opened 2021-06-18
Add option to skip tls verification

rikatz:
Signed-off-by: Ricardo Katz <rkatz@vmware.com>

madmin-go does not support skipping insecure verification.

This PR adds support for skipping the tls verification when using madmin-go client.

### DIFF
--- a/api.go
+++ b/api.go
@@ -91,8 +91,9 @@ const (
 
 // Options for New method
 type Options struct {
-	Creds  *credentials.Credentials
-	Secure bool
+	Creds              *credentials.Credentials
+	Secure             bool
+	InsecureSkipVerify bool
 	// Add future fields here
 }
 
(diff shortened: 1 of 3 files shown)


harshavardhana:
@rikatz this is not a correct fix, we allow you to override the Transport so during Transport initialization you need to provide the right TLSClientConfig

```go
        // API requests are secure (HTTPS) if secure=true and insecure (HTTPS) otherwise.
        // New returns an MinIO Admin client object.
        madmClnt, err := madmin.New("your-minio.example.com:9000", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
        if err != nil {
                log.Fatalln(err)
        }

        tr := &http.Transport{
                 TLSClientConfig:    &tls.Config{RootCAs: pool, InsecureSkipVerify: true},
                 DisableCompression: true,
        }
       madminClnt.SetCustomTransport(tr)
```

rikatz:
Oooh I've missed that one!! Thank you @harshavardhana 

closing here !

PR #13, opened 2021-06-29
Add NodeInfo interface

anjalshireesh:
With functions for getting and setting the address. This will be useful
in the upcoming feature of anonymizing addresses in the health report,
where a generic function can be written to update the anonymized address
on various structs that implement this interface.

### DIFF
--- a/health.go
+++ b/health.go
@@ -45,6 +45,12 @@ const (
 	HealthInfoVersion = HealthInfoVersion2
 )
 
+// NodeInfo - Interface to abstract any struct that contains address/endpoint and error fields
+type NodeInfo interface {
+	GetAddr() string
+	SetAddr(addr string) NodeInfo
+}
+
 // CPU contains system's CPU information.
 type CPU struct {
 	VendorID   string   `json:"vendor_id"`
@@ -66,6 +72,22 @@ type CPUs struct {
 	Error string `json:"error,omitempty"`
 
 	CPUs []CPU `json:"cpus,omitempty"`
+}
+
+func (c CPUs) GetAddr() string {
+	return c.Addr
+}
+
+func (c CPUs) SetAddr(addr string) NodeInfo {
+	c.Addr = addr
+	return c
+}
+
+func (c CPUs) Anonymize(addrAnonymizer func(string) (string, error)) CPUs {
+	newAddr, err := addrAnonymizer(c.Addr)
+	c.Addr = newAddr
+	c.Error = err.Error()
+	return c
 }
 
 // GetCPUs returns system's all CPU information.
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -538,6 +614,15 @@ type NetPerfInfo struct {
 	RemotePeers []PeerNetPerfInfo `json:"remote_peers,omitempty"`
 }
 
+func (n NetPerfInfo) GetAddr() string {
+	return n.Addr
+}
+
+func (n NetPerfInfo) SetAddr(addr string) NodeInfo {
+	n.Addr = addr
```
balamurugana, line 622: As `Addr` is available in all structures, you could have a base struct for `Addr` and extend in every structure. If `Addr` is `public`, there is no need to have `SetAddr()` method and you can set value directly.

anjalshireesh: @balamurugana I want to write a func in minio like:

```golang
anonymizeAddr := func(info madmin.NodeInfo) madmin.NodeInfo {
		newAddr := anonAddr(info.GetAddr())
		return info.SetAddr(newAddr)
	}
```
This can be used for any of the health structs that have an address field and implement the interface, thus avoiding repeating the similar code for each of the concerned structs.

Can something like this be done with what you are suggesting? Can you please help me with a code sample?

balamurugana:
```diff
diff --git a/health.go b/health.go
index bb8e761..d57e1ad 100644
--- a/health.go
+++ b/health.go
@@ -45,6 +45,24 @@ const (
 	HealthInfoVersion = HealthInfoVersion2
 )
 
+type NodeInfo interface {
+	GetAddr() string
+	SetAddr(addr string)
+}
+
+type NodeAddr struct {
+	Addr  string `json:"addr"`
+	Error string `json:"error,omitempty"`
+}
+
+func (node NodeAddr) GetAddr() string {
+	return node.Addr
+}
+
+func (node NodeAddr) SetAddr(addr string) {
+	node.Addr = addr
+}
+
 // CPU contains system's CPU information.
 type CPU struct {
 	VendorID   string   `json:"vendor_id"`
@@ -62,9 +80,7 @@ type CPU struct {
 
 // CPUs contains all CPU information of a node.
 type CPUs struct {
-	Addr  string `json:"addr"`
-	Error string `json:"error,omitempty"`
-
+	NodeAddr
 	CPUs []CPU `json:"cpus,omitempty"`
 }
 
@@ -73,8 +89,7 @@ func GetCPUs(ctx context.Context, addr string) CPUs {
 	infos, err := cpu.InfoWithContext(ctx)
 	if err != nil {
 		return CPUs{
-			Addr:  addr,
-			Error: err.Error(),
+			NodeAddr: NodeAddr{Addr: addr, Error: err.Error()},
 		}
 	}
 
@@ -107,7 +122,7 @@ func GetCPUs(ctx context.Context, addr string) CPUs {
 	}
 
 	return CPUs{
-		Addr: addr,
+		NodeAddr: NodeAddr{Addr: addr},
 		CPUs: cpus,
 	}
 }
```

Now you can do
```go
func redact(info health.NodeInfo) { info.SetAddr(anonAddr(info.GetAddr())) }
```
```go
redact(&healthInfo.Sys.CPUInfo[0])
```

anjalshireesh: Thanks @balamurugana , I've implemented according to your suggestion. PTAL.

balamurugana (approved):

harshavardhana (approved):

PR #14, opened 2021-07-08
Add Admin Inspect method

klauspost:
Add method for inspecting data.

### DIFF
--- a/inspect.go
+++ b/inspect.go
@@ -33,13 +33,13 @@ type InspectOptions struct {
 // Inspect makes an admin call to download a raw files from disk.
 func (adm *AdminClient) Inspect(ctx context.Context, d InspectOptions) (key [32]byte, c io.ReadCloser, err error) {
 	path := fmt.Sprintf(adminAPIPrefix + "/inspect-data")
+	q := make(url.Values)
+	q.Set("volume", d.Volume)
+	q.Set("file", d.File)
 	resp, err := adm.executeMethod(ctx,
 		http.MethodGet, requestData{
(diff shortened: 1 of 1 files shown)


inspect.go
```diff
@@ -0,0 +1,77 @@
...
+	Volume, File string
+}
+
+// Inspect makes an admin call to download a raw files from disk.
+func (adm *AdminClient) Inspect(ctx context.Context, d InspectOptions) (key [32]byte, c io.ReadCloser, err error) {
+	path := fmt.Sprintf(adminAPIPrefix + "/inspect-data")
+	resp, err := adm.executeMethod(ctx,
+		http.MethodGet, requestData{
+			relPath: path,
+			queryValues: url.Values{
```
vadmeste, line 39: Let's use golang API to work with url.Values,

```
q := make(url.Values)
q.Set("volume", d.Volume)
q.Set("file", d.File)
```

vadmeste (commented):
One comment

inspect.go
klauspost: Not sure I understand the difference, but no problem.

klauspost:
```
// Set sets the key to value. It replaces any existing
// values.
func (v Values) Set(key, value string) {
	v[key] = []string{value}
}
```
So yeah, the same AFAICT.

vadmeste: > So yeah, the same AFAICT.

yeah, it looks like the same :) I thought maybe Golang is doing something extra

harshavardhana (approved):

PR #15, opened 2021-07-13
Add Prefix Usage information in Account Info API 

vadmeste:
Return the recursive size of each first level prefix of given bucket in
the accounting info API.


vadmeste:
@harshavardhana PTAL

harshavardhana (approved):

PR #18, opened 2021-07-23
Add updatedb check in GetSysErrors

anjalshireesh:
Add a sys error if `updatedb` command is available in path.

Also improve the audit logging check to see if the `kauditd` process
is running, apart from checking for `audit=1` in /proc/cmdline

### DIFF
--- a/health.go
+++ b/health.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -280,9 +281,18 @@ func GetSysErrors(ctx context.Context, addr string) SysErrors {
 		se.Errors = append(se.Errors, SysErrAuditEnabled)
 	}
 
+	pkgCmd, arg := getPkgListCmd()
```
balamurugana, line 284: Use distro independent way to find `updatedb` provided by `mlocate` package in pure Go by
```go
if _, err := exec.LookPath("updatedb"); err == nil {
    return errors.New("updatedb is installed")
}
```

balamurugana (approved):

harshavardhana (approved):

PR #19, opened 2021-07-30
Add selinux check in sys errors api

anjalshireesh:
If the `getenforce` command is available, and it returns anything other
than `Disabled`, an error "selinux is not disabled" will be added to the
api response.

Signed-off-by: Shireesh Anjal <shireesh@minio.io>

### DIFF
--- a/health.go
+++ b/health.go
@@ -49,8 +49,9 @@ const (
 )
 
 const (
-	SysErrAuditEnabled      = "audit is enabled"
-	SysErrUpdatedbInstalled = "updatedb is installed"
+	SysErrAuditEnabled       = "audit is enabled"
+	SysErrUpdatedbInstalled  = "updatedb is installed"
+	SysErrSELinuxNotDisabled = "selinux is not disabled"
 )
 
 // NodeInfo - Interface to abstract any struct that contains address/endpoint and error fields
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -293,6 +294,12 @@ func GetSysErrors(ctx context.Context, addr string) SysErrors {
 		se.Errors = append(se.Errors, SysErrUpdatedbInstalled)
 	}
 
+	o, err := exec.Command("getenforce").Output()
```
balamurugana, line 297: There are many ways to check `SELinux`. My recommendation is to use `/etc/selinux/config` file.
1. If `/etc/selinux/config` is not found, `SELinux` is not installed.
2. If `/etc/selinux/config` contains a line `SELINUX=enforcing`, `SELinux` is enabled.
3. If `/etc/selinux/config` contains a line `SELINUX=permissive`, `SELinux` is running in warning mode i.e. it doesn't fail, but logs illegal access.
4. If `/etc/selinux/config` contains a line, `SELINUX=disabled`, `SELinux` is disabled.

Here, we shouldn't fail for (3), instead of raise a warning.

anjalshireesh: @harshavardhana are two levels (warning and error) really required here? The current implementation records anything that is not recommended as an error. Any entry in `SysErrors` will show up in red on the subnet health UI.

anjalshireesh:
It was discussed and decided that SELinux info will be added as a data point rather than an error. So closing this PR.

PR #20, opened 2021-08-09
Add function to get sys services

anjalshireesh:
The intention is to provide status of any sys services that can
potentially impact the performance of minio.

At present, it will return information about the `selinux` service
(not-installed/disabled/permissive/enforcing)

### DIFF
--- a/health.go
+++ b/health.go
@@ -309,33 +309,23 @@ func GetSysServices(ctx context.Context, addr string) SysServices {
 
 func getSELinuxInfo() (SysService, error) {
 	ss := SysService{Name: SrvSELinux}
-	configFile := "/etc/selinux/config"
 
-	if _, err := os.Stat(configFile); err != nil {
-		if os.IsNotExist(err) {
-			// selinux is not installed.
+	file, err := os.Open("/etc/selinux/config")
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
 			ss.Status = SrvNotInstalled
 			return ss, nil
-		} else {
-			return ss, err
 		}
-	}
-
-	file, err := os.Open(configFile)
-	if err != nil {
 		return ss, err
 	}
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -267,7 +291,58 @@ func GetOSInfo(ctx context.Context, addr string) OSInfo {
...
+			// selinux is not installed.
+			ss.Status = SrvNotInstalled
+			return ss, nil
+		} else {
+			return ss, err
+		}
+	}
+
+	file, err := os.Open(configFile)
+	if err != nil {
```
balamurugana, line 325: You could avoid `os.Stat()` and do the check like
```go
if errors.Is(err, os.ErrNotExist) {
    ss.Status = SrvNotInstalled
    return ss, nil
}
```

health.go
```diff
@@ -267,7 +291,58 @@ func GetOSInfo(ctx context.Context, addr string) OSInfo {
...
+		return ss, err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	statuses := []string{"enforcing", "permissive", "disabled"}
+	for scanner.Scan() {
+		txt := scanner.Text()
+		for _, status := range statuses {
+			if strings.Contains(txt, "SELINUX="+status) {
```
balamurugana, line 335: you could simplify this by
```go
tokens := strings.SplitN(strings.TrimSpace(scanner.Text()), "=", 2)
if len(tokens) == 2 && tokens[0] == "SELINUX" {
    ss.Status = tokens[1]
    return ss, nil
}
```

health.go
anjalshireesh: Done

health.go
anjalshireesh: Done

balamurugana (approved):

harshavardhana (approved):

PR #22, opened 2021-08-19
Sort the Speedtest() results

krishnasrinivas:


harshavardhana (changes requested):
Perhaps server should send a sorted list?

krishnasrinivas:
I think it's best to sort it in `mc` since mc is the one that needs it sorted.

PR #23, opened 2021-08-20
Add per storage class parity nubmer in Heal struct

vadmeste:
We are going to show disks/servers tolerance in `mc admin heal` status
output. For that we need the number of parity disks for each storage
class.

### DIFF
--- a/heal-commands.go
+++ b/heal-commands.go
@@ -309,15 +309,16 @@ type MRFStatus struct {
 
 // BgHealState represents the status of the background heal
 type BgHealState struct {
+	// Total items scanned by the continous background healing
 	ScannedItemsCount int64
-
+	// Disks currently in heal states
 	HealDisks []string
-
 	// SetStatus contains information for each set.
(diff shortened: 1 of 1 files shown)


heal-commands.go
```diff
@@ -309,15 +309,16 @@ type MRFStatus struct {
...
+	// Disks currently in heal states
 	HealDisks []string
-
 	// SetStatus contains information for each set.
 	Sets []SetStatus `json:"sets"`
-
 	// Endpoint -> MRF Status
 	MRF map[string]MRFStatus `json:"mrf"`
+	// Parity per storage class
+	SCParity map[string]int `json:"scParity"`
```
klauspost, line 321:
```suggestion
	SCParity map[string]int `json:"sc_parity"`
```
Snake case is used elsewhere and JSON default.

klauspost (changes requested):

klauspost (approved):

harshavardhana (approved):

PR #28, opened 2021-08-28
Support blank replies for speedtest API to prevent connection from disconnecting

krishnasrinivas:


speedtest.go
```diff
@@ -79,11 +78,15 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (Spee
 	if resp.StatusCode != http.StatusOK {
 		return SpeedTestResult{}, httpRespToErrorResponse(resp)
 	}
-	respBytes, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return SpeedTestResult{}, err
+	for {
+		var result SpeedTestResult
```
harshavardhana, line 82:
```suggestion
		var result SpeedTestResult
```

This needs to be out of `for {`

harshavardhana (changes requested):

harshavardhana (approved):

PR #30, opened 2021-09-15
Return chan for speedtest to support spinner in mc

krishnasrinivas:


speedtest.go
```diff
@@ -71,22 +71,24 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (Spee
...
+	go func() {
+		defer closeResponse(resp)
+		defer close(ch)
+		var result SpeedTestResult
+		for {
+			dec := json.NewDecoder(resp.Body)
+			if err := dec.Decode(&result); err != nil {
+				return
+			}
+			ch <- result
```
harshavardhana, line 90:
```suggestion
select {
			case ch <- result:
			case <-ctx.Done():
			}
```

harshavardhana (changes requested):

speedtest.go
krishnasrinivas: doing this can cause race condition between blank responses and the final response. i.e when the blank responses are being processed by `mc`, ctx.Done() might get closed causing `mc` to never read the last response

harshavardhana: @krishnasrinivas yeah but what is the point of reading last response? when you have context closed?

krishnasrinivas: for the simplest case, assume server sends a blank response and a valid response.
Now mc will read the blank response and processing it.
At this point madmin.SpeedTest() will be blocked trying to send the valid response onto the channel. And the resp.Body stream has reached the end. But since ctx.Done() gets closed we end up dropping the valid response since `mc` never consumed it

harshavardhana: Yeah but ctx.Done() only can happen if `mc` itself is about to die, that's why we should just use that here anyways.

harshavardhana: Do not see a reason why we would wait to consume something when our active context is going to die.

krishnasrinivas: oh no, I was incorrectly thinking about this

speedtest.go
```diff
@@ -71,22 +71,27 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (Spee
...
+		var result SpeedTestResult
+		for {
+			dec := json.NewDecoder(resp.Body)
+			if err := dec.Decode(&result); err != nil {
+				return
+			}
+			select {
+			case ch <- result:
+			case <-ctx.Done():
+			}
```
harshavardhana, line 93: in-fact we should return here, since we are done with this..

harshavardhana (changes requested):

harshavardhana (approved):

PR #32, opened 2021-09-27
Add support for cluster replication

donatello:


cluster-commands.go
```diff
@@ -0,0 +1,266 @@
...
+	return nil
+}
+
+// BktOp represents the bucket operation being requested.
+type BktOp string
+
+// BktOp value constants.
+const (
+	// make bucket and enable versioning
+	MakeWithVerBktOp BktOp = "make-with-ver"
```
vadmeste, line 164: No harm in using complete string 'make-with-version'

cluster-commands.go
```diff
@@ -0,0 +1,266 @@
...
+	Policy      string `json:"policy"`
+}
+
+// CRIAMItem - represents an IAM object that will be copied to a peer.
+type CRIAMItem struct {
+	Type          string            `json:"type"`
+	Name          string            `json:"name"`
+	Policy        *iampolicy.Policy `json:"policy"`
+	PolicyMapping *CRPolicyMapping  `json:"policyMapping"`
+	SvcAccChange  *CRSvcAccChange   `json:"serviceAccountChange"`
```
vadmeste, line 241: What is this ? Also maybe we can add IAM type (regular user, sts, svcacc), though I know we are going to support only service accounts for now but maybe not in the future

vadmeste (commented):
First round of review

cluster-commands.go
```diff
@@ -0,0 +1,266 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	iampolicy "github.com/minio/pkg/iam/policy"
+)
+
+// CRAdd - argument type for the CR add command.
+type CRAdd struct {
```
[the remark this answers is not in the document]
donatello: Made a fix, it is used in mc.

cluster-commands.go
donatello: 👍🏾

cluster-commands.go
donatello: `CRIAMItem` is a sort of "sum" type to represent different kinds of changes related to IAM - I am not sure if that's what you're asking. `SvcAccChange` just represents a service account change such as add/update/delete. If we support regular user, sts, etc we can add corresponding pointer type members to this struct.

donatello: Added some comments in the code, HTH.

harshavardhana (approved):

kannappanr (approved):
LGTM

go.mod
```diff
@@ -8,11 +8,10 @@ require (
 	github.com/minio/sha256-simd v1.0.0 // indirect
 	github.com/philhofer/fwd v1.1.1 // indirect
 	github.com/prometheus/procfs v0.7.3
+	github.com/minio/pkg v1.0.11
```
harshavardhana, line 11: @donatello you cannot import `minio/pkg` `minio/pkg` is AGPLv3 and `minio/madmin-go` is Apache they are incompatible for linkage i.e basically madmin-go would AGPLv3 which will lead to consumers of this SDK to be bounded by AGPLv3.

My suggestion is to avoid the usage and find replacements locally - madmin-go generally should be self-contained.

harshavardhana (changes requested):

cluster-commands.go
```diff
@@ -0,0 +1,357 @@
...
+	Policy      string `json:"policy"`
+}
+
+// CRIAMItem - represents an IAM object that will be copied to a peer.
+type CRIAMItem struct {
+	Type string `json:"type"`
+
+	// Name and Policy below are used when Type == CRIAMItemPolicy
+	Name   string            `json:"name"`
+	Policy *iampolicy.Policy `json:"policy"`
```
harshavardhana, line 273: json.RawMessage is appropriate here @donatello

harshavardhana (changes requested):

go.mod
donatello: Ah ok, let me fix this.

cluster-commands.go
donatello: > json.RawMessage is appropriate here @donatello

👍🏾

go.mod
donatello: Done!

harshavardhana (approved):

PR #34, opened 2021-10-01
procinfo: Do not treat username issue as an error

anjalshireesh:
If the system cannot fetch username of the minio process owner, it gets
reported as an error in the health report. However, this can happen for
legitimate reasons (e.g. in minio-operator based environments) and is
not something that needs to be flagged as an error.

Fixed by ignoring error when trying to get the username

### DIFF
--- a/health.go
+++ b/health.go
@@ -667,11 +667,9 @@ func GetProcInfo(ctx context.Context, addr string) ProcInfo {
 		return procInfo
 	}
 
-	procInfo.Username, err = proc.UsernameWithContext(ctx)
-	if err != nil {
-		procInfo.Error = err.Error()
-		return procInfo
-	}
+	// In certain environments, it is not possible to get username e.g. minio-operator
+	// Plus it's not a serious error. So ignore error if any.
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

health.go
```diff
@@ -667,11 +667,9 @@ func GetProcInfo(ctx context.Context, addr string) ProcInfo {
...
 	}
 
-	procInfo.Username, err = proc.UsernameWithContext(ctx)
-	if err != nil {
-		procInfo.Error = err.Error()
-		return procInfo
-	}
+	// In certain environments, it is not possible to get username e.g. minio-operator
+	// Plus it's not a serious error. So ignore error if any.
+	procInfo.Username, _ = proc.UsernameWithContext(ctx)
```
harshavardhana, line 672: If this fails we can add `procInfo.Username = "<non-root>"` as value instead of `empty` strings.

harshavardhana (changes requested):

health.go
anjalshireesh: Done

harshavardhana (approved):

PR #44, opened 2021-10-19
Add TLS info to minio info struct

anjalshireesh:
So that it can be captured as part of the health report.


health.go
```diff
@@ -790,6 +790,19 @@ type MinioInfo struct {
 	Services     Services     `json:"services,omitempty"`
 	Backend      interface{}  `json:"backend,omitempty"`
 	Servers      []ServerInfo `json:"servers,omitempty"`
+	TLS          TLSInfo      `json:"tls,omitempty"`
```
klauspost, line 793: Omitempty is pointless, since it is not a pointer.

```suggestion
	TLS          TLSInfo      `json:"tls"`
```

klauspost (approved):
LGTM

PR #48, opened 2021-11-15
Use pointer for TLS struct

anjalshireesh:
Instead of having a separate field `TLSInfoAvailable`, using pointer for
the `TLS` field serves the same purpose.


klauspost (approved):
LGTM.

harshavardhana (approved):

PR #53, opened 2021-12-11
Add cluster replication status command

poornas:


cluster-commands.go
```diff
@@ -359,6 +361,51 @@ func (adm *AdminClient) SRInternalReplicateBucketMeta(ctx context.Context, item
...
+		return info, httpRespToErrorResponse(resp)
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return info, err
+	}
+
+	err = json.Unmarshal(b, &info)
+	return info, err
```
harshavardhana, line 406:
```suggestion
        err = json.NewDecoder(resp.Body).Decode(&info)
        return info, err
```

cluster-commands.go
```diff
@@ -359,6 +361,51 @@ func (adm *AdminClient) SRInternalReplicateBucketMeta(ctx context.Context, item
...
+	// byte respresentation.
+	SSEConfig *string `json:"sseConfig,omitempty"`
+	// replication config in json representation
+	ReplicationConfig *string `json:"replicationConfig,omitempty"`
+}
+
+// SRInternalReplicateBucketMetaInfo - gets all bucket metadata from a peer cluster.
+func (adm *AdminClient) SRInternalReplicateBucketMetaInfo(ctx context.Context) (info SRBucketMetaInfo, err error) {
+	reqData := requestData{
+		relPath: adminAPIPrefix + "/site-replication/peer/bucket-meta",
```
harshavardhana, line 387: why do we need peer based calls here? - admin calls should be cluster wide not peer specific @poornas

cluster-commands.go
```diff
@@ -393,3 +440,133 @@ func (adm *AdminClient) SRInternalGetIDPSettings(ctx context.Context) (info IDPS
...
+	DeploymentID   string
+	Buckets        map[string]SRBucketMetaInfo   // map of bucket metadata info
+	Policies       map[string]json.RawMessage    //  map of IAM policy name to content
+	UserPolicies   map[string]SRPolicyMapping    // map of username -> user policy mapping
+	GroupPolicies  map[string]SRPolicyMapping    // map of groupname -> group policy mapping
+	ReplicationCfg map[string]replication.Config // map of bucket -> replication config
+}
+
+// SiteReplicationMetaInfo - returns site meta info for replication.
+func (adm *AdminClient) SiteReplicationMetaInfo(ctx context.Context) (info SRMetaInfo, err error) {
```
harshavardhana, line 457:
```suggestion
// ListSiteReplicationBuckets - returns site meta info for replication.
func (adm *AdminClient) ListSiteReplicationBuckets(ctx context.Context) (info SRBucketsResult, err error) {
```

cluster-commands.go
```diff
@@ -393,3 +440,133 @@ func (adm *AdminClient) SRInternalGetIDPSettings(ctx context.Context) (info IDPS
...
+	reqData := requestData{
+		relPath: adminAPIPrefix + "/site-replication/metainfo",
```
harshavardhana, line 459:
```suggestion
		relPath: adminAPIPrefix + "/site-replication/listbuckets",
```

cluster-commands.go
```diff
@@ -393,3 +440,133 @@ func (adm *AdminClient) SRInternalGetIDPSettings(ctx context.Context) (info IDPS
...
+		return info, httpRespToErrorResponse(resp)
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return info, err
+	}
+
+	err = json.Unmarshal(b, &info)
+	return info, err
```
harshavardhana, line 478:
```suggestion
        err = json.NewDecoder(resp.Body).Decode(&info)
        return info, err
```

harshavardhana (changes requested):

cluster-commands.go
harshavardhana: Okay then let's name it appropriately 

> SRInternalReplicateBucketMetaInfo 

to 

> SRRemoteReplicateBucketMetaInfo

harshavardhana: and also not call it "peer"

cluster-commands.go
harshavardhana: Then it should be perhaps `SRInfo`  not `SRMetaInfo` - and also `ListAllSites()`

harshavardhana (changes requested):
I think we should drop "Meta" from all the structs here its superfluous and not that meaningful. Instead it should be simply called as 

- SRInfo
- SRBucketInfo

Info itself says its some kind of "Metadata" -> "MetaInfo" is like saying "MetaMeta"

cluster-commands.go
```diff
@@ -393,3 +415,123 @@ func (adm *AdminClient) SRInternalGetIDPSettings(ctx context.Context) (info IDPS
...
+	DeploymentID   string
+	Buckets        map[string]SRBucketMetaInfo   // map of bucket metadata info
+	Policies       map[string]json.RawMessage    //  map of IAM policy name to content
+	UserPolicies   map[string]SRPolicyMapping    // map of username -> user policy mapping
+	GroupPolicies  map[string]SRPolicyMapping    // map of groupname -> group policy mapping
+	ReplicationCfg map[string]replication.Config // map of bucket -> replication config
+}
+
+// ListReplicationInfo - returns replication metadata info for a site.
+func (adm *AdminClient) ListReplicationInfo(ctx context.Context) (info SRInfo, err error) {
```
harshavardhana, line 432:
```suggestion
// GetSRInfo - returns replication metadata info for a site.
func (adm *AdminClient) GetSRInfo(ctx context.Context) (info SRInfo, err error) {
```

cluster-commands.go
```diff
@@ -359,6 +361,26 @@ func (adm *AdminClient) SRInternalReplicateBucketMeta(ctx context.Context, item
 	return nil
 }
 
+// SRBucketMetaInfo - returns all the bucket metadata available for bucket
+type SRBucketMetaInfo struct {
```
harshavardhana, line 365:
```suggestion
type SRBucketInfo struct {
```

harshavardhana (changes requested):
Also I think we already called "peer" for sites which is okay and its already there - lets not change the meaning now. 

We can do those API cleanups somewhere in the future.

harshavardhana (approved):

PR #55, opened 2021-12-16
Add IsKubernetes and IsDocker to MinioInfo

anjalshireesh:
So that this info can be added in the health info and health-analyzer
can use it to perform or skip certain checks.

### DIFF
--- a/health.go
+++ b/health.go
@@ -791,6 +791,8 @@ type MinioInfo struct {
 	Backend      interface{}  `json:"backend,omitempty"`
 	Servers      []ServerInfo `json:"servers,omitempty"`
 	TLS          *TLSInfo     `json:"tls"`
+	IsKubernetes bool         `json:"is_kubernetes"`
+	IsDocker     bool         `json:"is_docker"`
 }
 
 type TLSInfo struct {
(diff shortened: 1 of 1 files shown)


health.go
```diff
@@ -791,6 +791,8 @@ type MinioInfo struct {
 	Backend      interface{}  `json:"backend,omitempty"`
 	Servers      []ServerInfo `json:"servers,omitempty"`
 	TLS          *TLSInfo     `json:"tls"`
+	IsKubernetes bool         `json:"is_kubernetes"`
+	IsDocker     bool         `json:"is_docker"`
```
klauspost, line 795: Can this always be determined? Otherwise "false" may give a false indication (no pun intended), also for backwards compat all will return "false".

Maybe a `*bool` is more appropriate?

anjalshireesh: Planning to use the functions `IsKubernetes()` and `IsDocker()` from https://github.com/minio/minio/blob/master/cmd/update.go#L154 for populating these.

So I think they'll always be determined. @harshavardhana what do you think?

klauspost: @anjalshireesh Fields will not be present from older server versions. In that case they will be "false".

anjalshireesh: oh yes.. will change to `*bool`

klauspost (approved):

PR #61, opened 2022-01-06
Add types for OIDC + internal IDP support for site replication

donatello:


cluster-commands.go
```diff
@@ -261,9 +263,23 @@ type SRPolicyMapping struct {
...
+	SessionToken        string `json:"sessionToken"`
+	ParentUser          string `json:"parentUser"`
+	ParentPolicyMapping string `json:"parentPolicyMapping,omitempty"`
+}
+
+// SRIAMUser - represents a regular (IAM) user to be replicated. A nil UserReq
+// implies that a user delete operation should be replicated on the peer cluster.
+type SRIAMUser struct {
+	AccessKey string              `json:"accessKey"`
+	UserReq   *AddOrUpdateUserReq `json:"userReq"`
```
harshavardhana, line 277:
```suggestion
	UserReq   *AddOrUpdateUserReq `json:"userReq"`
	DeleteReq bool
```

Explicit key is better than using `nil` for UserReq to perform Delete calls.

cluster-commands.go
```diff
@@ -413,6 +462,14 @@ func (adm *AdminClient) SRPeerGetIDPSettings(ctx context.Context) (info IDPSetti
...
 
 	err = json.Unmarshal(b, &info)
+	if err != nil {
+		// If the server is older version, the IDPSettings was =
+		// LDAPSettings, so we try that.
+		err2 := json.Unmarshal(b, &info.LDAP)
+		if err2 == nil {
+			err = nil
+		}
+	}
```
harshavardhana, line 472: @donatello perhaps you can write a single UnmarshalJSON instead to support both cases instead of this kind of fallback. This way the struct has the right method for picking up the right struct.

thoughts?

harshavardhana (changes requested):

cluster-commands.go
donatello: Not sure if it's a good idea - we only need the fallback unmarshal behavior in the specific case of supporting an older version server - there could be an issue in the future if `IDPSettings` is used in some other place.

harshavardhana: Okay @donatello

harshavardhana (approved):

PR #62, opened 2022-01-06
add gofumpt cleanup

harshavardhana:


donatello:
@harshavardhana It might be simpler to just enable gofumpt in golangci-lint -> https://golangci-lint.run/usage/configuration/#config-file

It doesn't have to be installed as golangci-lint comes with it

donatello:
CI and local development would be slightly faster :P

donatello (approved):

PR #63, opened 2022-01-13
Change XL Disk API latency to integer

vadmeste:
Having the latency in string does not help consuming by another code.
Modify it to integer, duration in nanosecond.

We need to release minio then release mc with madmin-go after this is getting merged.

### DIFF
--- a/info-commands.go
+++ b/info-commands.go
@@ -331,9 +331,9 @@ type ServerProperties struct {
 
 // DiskMetrics has the information about XL Storage APIs
 // the number of calls of each API and the moving average of
-// the duration of each API.
+// the duration, in nanosecond, of each API.
 type DiskMetrics struct {
-	APILatencies map[string]string `json:"apiLatencies,omitempty"`
+	APILatencies map[string]uint64 `json:"apiLatencies,omitempty"`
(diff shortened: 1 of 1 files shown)


info-commands.go
```diff
@@ -331,9 +331,9 @@ type ServerProperties struct {
 
 // DiskMetrics has the information about XL Storage APIs
 // the number of calls of each API and the moving average of
-// the duration of each API.
+// the duration, in nanosecond, of each API.
 type DiskMetrics struct {
-	APILatencies map[string]string `json:"apiLatencies,omitempty"`
+	APILatencies map[string]uint64 `json:"apiLatencies,omitempty"`
```
harshavardhana, line 336:
```suggestion
	APILatencies map[string]interface{} `json:"apiLatencies,omitempty"`
```

And convert it in `mc` @vadmeste

harshavardhana (changes requested):

info-commands.go
vadmeste: > And convert it in `mc` @vadmeste

yes, good idea!

harshavardhana (approved):

PR #64, opened 2022-01-14
Add cluster replication remove API commands

poornas:


cluster-commands.go
```diff
@@ -593,3 +593,74 @@ func (adm *AdminClient) SRStatusInfo(ctx context.Context) (info SRStatusInfo, er
...
+}
+
+// SRRemoveReq - arg body for SRRemoveReq
+type SRRemoveReq struct {
+	SiteNames []string `json:"sites"`
+	RemoveAll bool     `json:"all"` // true if all sites are to be removed.
+}
+
+const (
+	ReplicateRemoveStatusSuccess = "Requested sites were removed from cluster replication successfully."
```
harshavardhana, line 664:
```suggestion
	ReplicateRemoveStatusSuccess = "Requested site(s) were removed from cluster replication successfully."
```

cluster-commands.go
```diff
@@ -593,3 +593,74 @@ func (adm *AdminClient) SRStatusInfo(ctx context.Context) (info SRStatusInfo, er
...
+	ReplicateRemoveStatusPartial = "Some sites could not be removed from cluster replication configuration."
```
harshavardhana, line 665:
```suggestion
	ReplicateRemoveStatusPartial = "Some site(s) could not be removed from cluster replication configuration."
```

cluster-commands.go
harshavardhana, line 664: Are these const messages supposed to be in this library? @poornas I don't see them getting used anywhere. If this is `mc` specific - please move them to `mc`.

cluster-commands.go
```diff
@@ -593,3 +593,74 @@ func (adm *AdminClient) SRStatusInfo(ctx context.Context) (info SRStatusInfo, er
...
+	if err != nil {
+		return st, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return st, httpRespToErrorResponse(resp)
+	}
+	return ReplicateRemoveStatus{
+		Success: true,
+	}, nil
```
harshavardhana, line 620: `Success: true` is frivolous since we return err as `nil` so whatever is returned here is a success.

cluster-commands.go
```diff
@@ -593,3 +593,74 @@ func (adm *AdminClient) SRStatusInfo(ctx context.Context) (info SRStatusInfo, er
...
```
harshavardhana, line 647: are you adding this to keep this future-proof?

cluster-commands.go
```diff
@@ -593,3 +593,74 @@ func (adm *AdminClient) SRStatusInfo(ctx context.Context) (info SRStatusInfo, er
...
+}
+
+// ReplicateRemoveStatus - returns status of unlink request.
+type ReplicateRemoveStatus struct {
+	Success   bool   `json:"success"`
+	Status    string `json:"status"`
+	ErrDetail string `json:"errorDetail,omitempty"`
+}
```
harshavardhana, line 655: you only need 'Status` and `ErrDetail` - also do not see this data structure being used anywhere -is the server supposed to send this? - looks like the API methods are missing `json.Unmarshal`

harshavardhana (changes requested):

harshavardhana (approved):

cluster-commands.go
poornas: used in response from peer clusters as well as in mc

cluster-commands.go
poornas: removed this

cluster-commands.go
poornas: Removed Success; peer clusters and server use this struct

PR #65, opened 2022-01-14
API for netperf

krishnasrinivas:


mperf.go
```diff
@@ -0,0 +1,58 @@
...
+		})
+	if err != nil {
+		return result, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return result, httpRespToErrorResponse(resp)
+	}
+	err = json.NewDecoder(resp.Body).Decode(&result)
+	return result, err
+}
```
harshavardhana, line 58: I think its better to do it like an option for `speedtest` @krishnasrinivas not as a new API

harshavardhana (changes requested):

mperf.go
krishnasrinivas: Doing it as one API will make the API complicated @harshavardhana for ex. the response structure is different for mperf, dperf and object tests

harshavardhana: No it needs to be a new SpeedTestV2 API that replies back a data structure like this 

```
type SpeedtestResult struct {
         NetworkInfo
         DiskInfo
         ObjectLayerInfo
}
```

This provides a more consistent view from `mc` - separate APIs are not needed IMO.  This is in line with how subnet health does things.

harshavardhana (approved):

PR #70, opened 2022-02-02
Add frequency governor info to CPUs struct

anjalshireesh:
Adding a new bool pointer field `IsFreqGovPerf` under `CPUs`

It's value will be:
 - `nil` if frequency governor data couldn't be fetched (e.g. non-linux platforms)
 - `true` the governor is set to `performance` on all cores
 - `false` - if it is not set to `performance` on at least one core

This data can later be consumed by `health-analyzer` to add a health check on SUBNET


balamurugana (approved):

cpu_linux.go
```diff
@@ -0,0 +1,39 @@
...
+		return "", err
+	}
+
+	stats, err := fs.SystemCpufreq()
+	if err != nil {
+		return "", err
+	}
+
+	if len(stats) > 0 {
+		return stats[0].Governor, nil // assuming its same on all cpus/cores
```
vadmeste, line 36: https://www.kernel.org/doc/html/latest/admin-guide/pm/cpufreq.html#generic-scaling-governors

I think for Linux, we can simply check if all cores have 'performance' - if not return any other value

anjalshireesh: Are you suggesting having a `bool` field called `IsGovernorPerformance` instead of a `string` field that returns the configured value?

vadmeste: > Are you suggesting having a `bool` field called `IsGovernorPerformance` instead of a `string` field that returns the configured value?

No, my idea is to check for all cores if they are performance or not and returning whatever value which is not performance

vadmeste:
```
governor := "performance"
for _, st := range stats {
  if st.Governor != "performance" {
    governor = st.Govoernor
  }
}

anjalshireesh: As discussed, have implemented a `bool` based solution. PTAL.

vadmeste (approved):
LGTM, not tested

PR #76, opened 2022-03-22
Add types to include perf results in diag response

anjalshireesh:
For including the new perf test (speedtest) results in the output of `mc support diag`,
- Move the current perf related types (`PerfInfo` and others) to `health-old.go`
- Move the current `HealthInfo` struct also to `health-old.go` and rename it to `HealthInfoV2`
- Introduce new type `SpeedTestResults` to hold all speedtest results
- Introduce new `HealthInfo` which is same as the current one, except for using `SpeedTestResults` instead of `PerfInfo`
- Change names of perf tests So that when this mc is used with older versions of minio, it doesn't request for the old performance tests that are slow and are not very reliable.
- Bump up the health info version

### DIFF
--- a/health-old.go
+++ b/health-old.go
@@ -48,6 +48,41 @@ type HealthInfoV2 struct {
 	Minio     MinioHealthInfo `json:"minio,omitempty"`
 }
 
+func (info HealthInfoV2) String() string {
+	data, err := json.Marshal(info)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


kannappanr (approved):
LGTM

kannappanr:
ping @balamurugana

balamurugana (approved):

PR #77, opened 2022-03-31
Change support profile API

poornas:
Remove DownloadingProfile API, and merging functionality into
Profile API call.

### DIFF
--- a/profiling-commands.go
+++ b/profiling-commands.go
@@ -52,7 +52,7 @@ type StartProfilingResult struct {
 	Error    string `json:"error"`
 }
 
-// StartProfiling makes an admin call to remotely start profiling on a standalone
+// Deprecated: StartProfiling makes an admin call to remotely start profiling on a standalone
 // server or the whole cluster in  case of a distributed setup. [DEPRECATED]
 func (adm *AdminClient) StartProfiling(ctx context.Context, profiler ProfilerType) ([]StartProfilingResult, error) {
 	v := url.Values{}
--- a/profiling-commands.go
+++ b/profiling-commands.go
@@ -86,7 +86,7 @@ func (adm *AdminClient) StartProfiling(ctx context.Context, profiler ProfilerTyp
 	return startResults, nil
 }
 
-// DownloadProfilingData makes an admin call to download profiling data of a standalone
+// Deprecated: DownloadProfilingData makes an admin call to download profiling data of a standalone
 // server or of the whole cluster in  case of a distributed setup.[DEPRECATED]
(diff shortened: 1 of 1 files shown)


vadmeste (approved):
LGTM

profiling-commands.go
```diff
@@ -85,8 +86,8 @@ func (adm *AdminClient) StartProfiling(ctx context.Context, profiler ProfilerTyp
 	return startResults, nil
 }
 
-// DownloadProfilingData makes an admin call to download profiling data of a standalone
-// server or of the whole cluster in  case of a distributed setup.
+//  DownloadProfilingData makes an admin call to download profiling data of a standalone
```
harshavardhana, line 89:
```suggestion
// DownloadProfilingData makes an admin call to download profiling data of a standalone
```

harshavardhana (changes requested):

profiling-commands.go
```diff
@@ -52,7 +53,7 @@ type StartProfilingResult struct {
 }
 
 // StartProfiling makes an admin call to remotely start profiling on a standalone
```
harshavardhana, line 55:
```suggestion
// Deprecated: StartProfiling makes an admin call to remotely start profiling on a standalone
```

profiling-commands.go
```diff
@@ -52,7 +53,7 @@ type StartProfilingResult struct {
...
-// server or the whole cluster in  case of a distributed setup.
+// server or the whole cluster in  case of a distributed setup. [DEPRECATED]
```
harshavardhana, line 56:
```suggestion
// server or the whole cluster in case of a distributed setup.
```

profiling-commands.go
```diff
@@ -86,7 +87,7 @@ func (adm *AdminClient) StartProfiling(ctx context.Context, profiler ProfilerTyp
 }
 
 // DownloadProfilingData makes an admin call to download profiling data of a standalone
```
harshavardhana, line 89:
```suggestion
// Deprecated: DownloadProfilingData makes an admin call to download profiling data of a standalone
```

profiling-commands.go
```diff
@@ -86,7 +87,7 @@ func (adm *AdminClient) StartProfiling(ctx context.Context, profiler ProfilerTyp
...
-// server or of the whole cluster in  case of a distributed setup.
+// server or of the whole cluster in  case of a distributed setup.[DEPRECATED]
```
harshavardhana, line 90:
```suggestion
// server or of the whole cluster in case of a distributed setup.
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #81, opened 2022-05-04
Add struct for Kubernetes platform info

anjalshireesh:
This will be populated and added in the health data
(output of `mc support diag` command)


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #82, opened 2022-05-16
Add admin bucket import/export API calls

poornas:


bucket-metadata.go
```diff
@@ -0,0 +1,83 @@
...
+
+	resp, err := adm.executeMethod(ctx,
+		http.MethodPut, requestData{
+			relPath:     path,
+			queryValues: queryValues,
+			content:     content,
+		},
+	)
+	if err != nil {
+		closeResponse(resp)
```
harshavardhana, line 70: You can just do `defer closeResponse(resp)`  you are leaking this in the Good case.

bucket-metadata.go
```diff
@@ -0,0 +1,83 @@
...
+			queryValues: queryValues,
+		},
+	)
+	if err != nil {
+		closeResponse(resp)
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp)
```
harshavardhana, line 46: leaking resp.Body here while returning error

bucket-metadata.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	if resp.Body == nil {
```
harshavardhana, line 49: resp.Body is never nil and also leaking resp.Body here while returning error

harshavardhana (changes requested):

harshavardhana (changes requested):

harshavardhana (approved):

PR #83, opened 2022-06-01
Add admin iam import/export API calls

poornas:


examples/iam-migrate.go
```diff
@@ -0,0 +1,53 @@
...
+	// API requests are secure (HTTPS) if secure=true and insecure (HTTP) otherwise.
+	// New returns an MinIO Admin client object.
+	madmClnt, err := madmin.New("your-minio.example.com:9000", "YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", true)
+	if err != nil {
+		log.Fatalln(err)
+	}
+	ctx := context.Background()
+	// return IAM info as zipped content
+	r, ierr := client.ExportIAM(context.Background())
+	log.Fatalln(err)
```
harshavardhana, line 43: Please fix the examples @poornas they look incomplete..

iam-migrate.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	if resp.Body == nil {
+		return nil, errors.New("body is nil")
+	}
```
harshavardhana, line 47: resp.Body is leaking upon returns here..

iam-migrate.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return httpRespToErrorResponse(resp)
+	}
+
+	if resp.Body == nil {
+		return errors.New("body is nil")
+	}
+
```
harshavardhana, line 74: use `defer closeResponse(resp)`  

closeResponse is not needed when executeMethod() actually fails since that failure is when resp is `nil`.  It is when resp.StatusCode is not 200 OK where resp is non-nil and needs to be closed.

Not sure where you are copying this code from, we should fix that code @poornas

harshavardhana (changes requested):

examples/iam-migrate.go
harshavardhana: @poornas this example fails right here even when `err == nil` @poornas

poornas: fixed

iam-migrate.go
```diff
@@ -0,0 +1,70 @@
...
+	if resp.StatusCode != http.StatusOK {
+		closeResponse(resp)
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	if resp.Body == nil {
+		closeResponse(resp)
+		return nil, errors.New("body is nil")
+	}
```
harshavardhana, line 48: resp.Body is never nil this code is not needed.

iam-migrate.go
```diff
@@ -0,0 +1,70 @@
...
+	return resp.Body, nil
+}
+
+// ImportIAM makes an admin call to setup IAM  from imported content
+func (adm *AdminClient) ImportIAM(ctx context.Context, content []byte) error {
```
harshavardhana, line 53: how about we take a `contentReader` instead of []byte stream? - since we are return a io.ReadCloser with Export API

harshavardhana (changes requested):

harshavardhana (approved):

iam-migrate.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+)
+
+// ExportIAM makes an admin call to export IAM data
+func (adm *AdminClient) ExportIAM(ctx context.Context) (io.ReadCloser, error) {
+	path := fmt.Sprintf(adminAPIPrefix + "/export-iam")
```
harshavardhana, line 29:
```suggestion
	path :=adminAPIPrefix + "/export-iam"
```

iam-migrate.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// ImportIAM makes an admin call to setup IAM  from imported content
+func (adm *AdminClient) ImportIAM(ctx context.Context, contentReader io.ReadCloser) error {
+	content, err := ioutil.ReadAll(contentReader)
+	if err != nil {
+		return err
+	}
+
+	path := fmt.Sprintf(adminAPIPrefix + "/import-iam")
```
harshavardhana, line 54:
```suggestion
	path := adminAPIPrefix + "/import-iam"
```

harshavardhana (changes requested):
seems simpler not sure why you would need fmt.Sprintf() - everything LGTM

harshavardhana (approved):

PR #89, opened 2022-06-15
Add an option to set custom transport for Healthy() function

Praveenrajmani:


vadmeste (approved):
LGTM

Praveenrajmani:
@harshavardhana Can we export  `NewAnonymousClient` function and use them from outside like other clients? this will make it easier for http client level customizations

harshavardhana:
> @vadmeste @harshavardhana Can we export `NewAnonymousClient` function and use them from outside like other clients? this will make it easier for http client level customizations

Sure

PR #91, opened 2022-06-21
Add metrics and clean up trace

klauspost:
Add realtime metrics structs and add client that can be used for several metric types. Currently only scanner.

Refactors Trace options. Backwards compatible with older servers, but breaks API (struct changes).
API breakage is to make `TraceInfo` fields optional for different types.

PR for minio, console & mc are ready and will be added when this is in.


krisis (approved):

metrics.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// Contains returns whether m contains all of x.
+func (m MetricType) Contains(x MetricType) bool {
+	return m&x == x
+}
+
+// RealtimeMetrics provides realtime metrics.
+// This is intended to be expanded over time to cover more types.
+type RealtimeMetrics struct {
```
vadmeste, line 106: maybe AggregatedMetrics seems to be a better name here

vadmeste (commented):
Console UI has tracing as well, so we need to make sure that we upgrading madmin-go should be accompanied by console fixes (if I am not mistaken, type of tracing parameters

metrics.go
klauspost: I think `RealtimeMetrics` is more specific. I will take inspiration and change the field "Total" to "Aggregated".

klauspost:
> Console UI has tracing as well, so we need to make sure that we upgrading madmin-go should be accompanied by console fixes (if I am not mistaken, type of tracing parameters

@vadmeste Console PR: https://github.com/minio/console/pull/2141

info-commands.go
```diff
@@ -341,8 +341,8 @@ type ServerProperties struct {
 // the number of calls of each API and the moving average of
 // the duration, in nanosecond, of each API.
 type DiskMetrics struct {
-	APILatencies map[string]interface{} `json:"apiLatencies,omitempty"`
-	APICalls     map[string]uint64      `json:"apiCalls,omitempty"`
+	LastMinute map[string]TimedAction `json:"last_minute,omitempty"`
+	APICalls   map[string]uint64      `json:"apiCalls,omitempty"`
 }
```
harshavardhana, line 346: There are some breaking changes here @klauspost - it is going to be tricky to perhaps managing the client versions and console UI versions.

harshavardhana (changes requested):

info-commands.go
harshavardhana: If we do this with semantic versioning we should move madmin-go to v2.0.0

klauspost: @harshavardhana There is no breakage on Console. It does not use `APILatencies`.

For `mc` the changes are compatible:

Old mc, New MinIO -> No `APILatencies` are shown in `mc admin info --json`, but everything otherwise works. Solution is to upgrade `mc`.

New mc, Old MinIO -> No `LastMinute ` are shown in `mc admin info --json`, but everything otherwise works. Solution is to downgrade `mc` (keep it at server level) or upgrade MinIO server.

So nothing stops working.

harshavardhana (changes requested):

service-commands.go
```diff
@@ -174,46 +174,48 @@ func (adm AdminClient) ServiceTrace(ctx context.Context, opts ServiceTraceOpts)
 
 			dec := json.NewDecoder(resp.Body)
 			for {
-				var info TraceInfo
+				var info traceInfoLegacy
 				if err = dec.Decode(&info); err != nil {
-					closeResponse(resp)
```
harshavardhana, line 145: do we not want to close the response upon error? @klauspost

harshavardhana (changes requested):

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #96, opened 2022-07-05
Fix legacy trace conversion

klauspost:
Legacy types were not converted correctly, missing details.

Affects `mc admin trace`, `mc admin top api`.

Regression from #91


klauspost:
(forgot that old struct didn't have `omitempty`, so all fields were sent even if irrelevant)

harshavardhana (approved):

PR #99, opened 2022-07-22
Add Limit field to MemInfo

anjalshireesh:
So that it gets added in the health report and can be shown in the
SUBNET UI.

This will return the cgroup memory limit when configured (and less than
the system RAM), and the system RAM limit otherwise.

Most of the code related to getting the cgroup limit has been taken from
https://github.com/minio/pkg/tree/main/cgroup

### DIFF
--- a/cgroup/linux.go
+++ b/cgroup/linux.go
@@ -1,22 +1,21 @@
 //go:build linux
 // +build linux
 
-// Copyright (c) 2015-2022 MinIO, Inc.
 //
-// This file is part of MinIO Object Storage stack
+// MinIO Object Storage (c) 2022 MinIO, Inc.
 //
-// This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU Affero General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


cgroup/linux.go
```diff
@@ -0,0 +1,178 @@
...
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
```
harshavardhana, line 18: You need to change the license for this project, it can't be agpl

harshavardhana (changes requested):

cgroup/linux.go
anjalshireesh: Thanks, done.

harshavardhana (approved):

PR #101, opened 2022-07-22
add `IsEncrypted` function

aead:
This commit adds `IsEncrypted` which
reports whether data is encrypted
using the `EncryptData` function.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #102, opened 2022-08-02
Return the status of node configured for anonymous client in Alive

sinhaashish:
Return the status of the node that receives the request if there is no specified servers list.
This change is done for `mc ping ` command to return the status of the node which is configured as alias , when an empty list is passed to Alive() function.

Refer https://github.com/minio/mc/pull/4163#discussion_r934289349

### DIFF
--- a/cluster-health.go
+++ b/cluster-health.go
@@ -142,17 +142,17 @@ func (an *AnonymousClient) Alive(ctx context.Context, opts AliveOpts, servers ..
 	go func() {
 		defer close(resultsCh)
 		if len(servers) == 0 {
-			an.fetchData(ctx, an.endpointURL, resource, resultsCh)
+			an.alive(ctx, an.endpointURL, resource, resultsCh)
 		} else {
 			for _, server := range servers {
 				u, err := url.Parse(an.endpointURL.Scheme + "://" + server.Endpoint)
-				an.fetchData(ctx, u, resource, resultsCh)
 				if err != nil {
 					resultsCh <- AliveResult{
 						Error: err,
 					}
 					return
 				}
+				an.alive(ctx, u, resource, resultsCh)
 
 			}
 		}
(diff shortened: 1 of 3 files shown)


cluster-health.go
```diff
@@ -141,44 +141,53 @@ func (an *AnonymousClient) Alive(ctx context.Context, opts AliveOpts, servers ..
...
+			Online:       resp.StatusCode == http.StatusOK && resp.Header.Get("x-minio-server-status") != "offline",
+		}
+	}
+
+	select {
+	case <-ctx.Done():
+		return
+	case resultsCh <- result:
+	}
+}
```
harshavardhana, line 193: This function name should be called `alive()`

cluster-health.go
```diff
@@ -141,44 +141,53 @@ func (an *AnonymousClient) Alive(ctx context.Context, opts AliveOpts, servers ..
...
-			} else {
-				result = AliveResult{
-					Endpoint:     u,
-					ResponseTime: responseTime,
-					Online:       resp.StatusCode == http.StatusOK && resp.Header.Get("x-minio-server-status") != "offline",
+		if len(servers) == 0 {
+			an.fetchData(ctx, an.endpointURL, resource, resultsCh)
+		} else {
+			for _, server := range servers {
+				u, err := url.Parse(an.endpointURL.Scheme + "://" + server.Endpoint)
```
harshavardhana, line 148: do not ignore errors before calling `an.fetchData`

harshavardhana (changes requested):

harshavardhana (approved):

cluster-health.go
```diff
@@ -141,44 +141,53 @@ func (an *AnonymousClient) Alive(ctx context.Context, opts AliveOpts, servers ..
...
+		if len(servers) == 0 {
+			an.alive(ctx, an.endpointURL, resource, resultsCh)
+		} else {
+			for _, server := range servers {
+				u, err := url.Parse(an.endpointURL.Scheme + "://" + server.Endpoint)
+				if err != nil {
+					resultsCh <- AliveResult{
+						Error: err,
+					}
+					return
```
vadmeste, line 153: Shall we continue here instead of returning ?

vadmeste (commented):
One comment, LGTM otherwise

cluster-health.go
sinhaashish: done.

harshavardhana (approved):

PR #104, opened 2022-08-04
Inspect to return a new format

vadmeste:
Return format version in addition to the existing API to
interpret the inspect data correctly.

### DIFF
--- a/inspect.go
+++ b/inspect.go
@@ -18,6 +18,7 @@ package madmin
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


inspect.go
```diff
@@ -31,7 +31,7 @@ type InspectOptions struct {
 }
 
 // Inspect makes an admin call to download a raw files from disk.
-func (adm *AdminClient) Inspect(ctx context.Context, d InspectOptions) (key [32]byte, c io.ReadCloser, err error) {
+func (adm *AdminClient) Inspect(ctx context.Context, d InspectOptions) (format byte, key []byte, c io.ReadCloser, err error) {
```
harshavardhana, line 34: @vadmeste Let's return something called as `InspectReader`  that automatically 

- decrypts if key is present
- detects which format the inspect data is

We have no reason to expose this to the caller - caller should simply `io.ReadAll(ird)`  IMO.

harshavardhana (changes requested):

inspect.go
vadmeste: @harshavardhana we actually need to return a zip file which contains a plaintext file and some encrypted data. The zip file will be sent by email or uploaded in a portal :)

The zip file needs to be stored in a disk first, because decompressing it requires a seek feature io.ReaderAt

Now, why the server generates zip file in the first place? because the size of files inside it are stored after the data itself and this is needed so we can feed the zip file in a streaming fashion.

So overall InspectReader to decrypt on the fly is not possible.

harshavardhana: Again all of that is possible using a single reader @vadmeste currently functionality of replying key and all barebones is complicated.

The logic of extracting plaintext, and encrypted must be in this SDK alone - caller implementations can diverge and can lead to bugs. 

You return something called as `*inspectReader` now this reader if you do `io.Copy()` gives the encrypted content however if you do `(ir *InspectReader) ClusterInfo() Info` returns the relevant info.

vadmeste: > Again all of that is possible using a single reader @vadmeste 

This is only possible if we save the data in the disk, which is probably not very elegant as an SDK API.

I think let's do another round of discussion, maybe we don't need all of this.

harshavardhana: How you can keep them in memory @vadmeste ?

vadmeste: > How? - you can keep them in memory @vadmeste ?

Keep the whole inspect data in the memory ? the inspect data can be huge

harshavardhana: Not whole just the one that unencrypted.

harshavardhana: The current implementation is very low level we have to fix it because this looks quite strange implementation wise.

As it expects mc to do all the work by returning the entire skeleton of the protocol.

We should never have callers do so much work IMO, we have to come up with a simple API that returns reader.

This reader may have additional things so it can be a InspectReader that allows manipulating the stream but never the way it is implemented in the PR.

We are expecting callers to have their own implementations of interpreting the returned values here. 

This leads to diverged implementations and future breakage.

inspect.go
```diff
@@ -28,43 +29,68 @@ import (
...
+	}
+
+	method := ""
+	reqData := requestData{
+		relPath: fmt.Sprintf(adminAPIPrefix + "/inspect-data"),
+	}
+
+	// If the public-key is specified, create a POST request and send
+	// parameters as multipart-form instead of query values
+	if d.PublicKey != nil {
```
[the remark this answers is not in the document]
vadmeste: > You don't need it to be body, it simply needs to be GET with a http header and it's base64 value. Just like SSE-C style you basically ask for to be encrypted via this public key.

There is a limit for the header value in HTTP, but sometimes, some set a max value, like 4k for nginx.. so I thought POST form is a better way to send the request.. my testing public key is only 1.9 kb.. but I am not sure if we can have larger than that in the future (maybe we can embed multiple certificates in the future)..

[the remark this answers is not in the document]
harshavardhana: okay can you rebase this? and let me know if this is ready I can start testing it.

harshavardhana (changes requested):
I see some discussions still on this?, is this finalized? @vadmeste 

// cc @klauspost

vadmeste:
Not needed anymore, superseded by https://github.com/minio/madmin-go/pull/133/

PR #106, opened 2022-08-05
Add types for admin-rebalance subcommands

krisis:
.. and for tracing rebalance internal functions.

### DIFF
--- a/rebalance.go
+++ b/rebalance.go
@@ -1,0 +1,129 @@
+//
+// MinIO Object Storage (c) 2022 MinIO, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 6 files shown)


harshavardhana (approved):

rebalance.go
```diff
@@ -0,0 +1,178 @@
...
+		PercentFreeGoal: rm.PercentFreeGoal,
+		PoolStats:       rm.PoolStats,
+		StoppedAt:       rm.StoppedAt,
+	}
+	return json.Marshal(r)
+}
+
+func (rm *RebalanceMeta) UnmarshalJSON(b []byte) error {
+	type rebalMeta struct {
+		ARN             uuid.UUID
```
vadmeste, line 85: This doesn't have json tags like MarshalJSON(), it doesn't seem this will work correctly

vadmeste (commented):
One comment

rebalance.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return arn, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return arn, httpRespToErrorResponse(resp)
+	}
+
+	var rebalInfo struct {
+		ARN uuid.UUID `json:"arn"`
```
harshavardhana, line 115: ARN has a specific meaning in AWS S3 spec not sure it would be wise to use UUID as ARN - you can just call it as ID @krisis

harshavardhana (changes requested):

harshavardhana (approved):

harshavardhana:
Please rebase the PR @krisis

harshavardhana:
v1.6.1 is pushed @krisis FYI.

PR #108, opened 2022-08-08
Prevent divide by zero error on empty slice

anjalshireesh:
TimeDurations.Measure() is failing with divide by zero error when there
are no elements in the slice. This happens when generating health report
on a standalone server setup.

Fixed by returning struct with zero values in this scenario.


harshavardhana (approved):

vadmeste (approved):
LGTM

PR #127, opened 2022-09-23
Add AccountInfo setting

jy-kkkim:
The old AccountInfo example file does not work properly.
It has been modified with an example file suitable for the current version.


vadmeste (approved):
LGTM

PR #128, opened 2022-09-23
Add Example ServiceAccount

jy-kkkim:
Added a new example of the ServiceAccount function.

### DIFF
--- a/examples/service-account.go
+++ b/examples/service-account.go
@@ -1,0 +1,70 @@
+//go:build ignore
+// +build ignore
+//
+// MinIO Object Storage (c) 2022 MinIO, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


examples/service-account.go
```diff
@@ -0,0 +1,68 @@
+//
+// MinIO Object Storage (c) 2021 MinIO, Inc.
```
vadmeste, line 2: s/2021/2022

examples/service-account.go
```diff
@@ -0,0 +1,68 @@
+//
```
vadmeste, line 1: You need to add this
```
 //go:build ignore
 // +build ignore
```

so this example won't be compiled with go lint and others

examples/service-account.go
vadmeste, line 2: s/2021/2022

harshavardhana (approved):

vadmeste (approved):
LGTM

PR #131, opened 2022-09-29
Use new KMS handlers instead of old ones that use admin:* actions

reivaj05:


harshavardhana (changes requested):
you can add Deprecated for older API calls refer https://go.dev/blog/godoc

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary, but must be kept for compatibility with existing programs. To signal that an identifier should not be used, add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

kms-commands.go
```diff
@@ -129,6 +150,21 @@ func (adm *AdminClient) CreateKey(ctx context.Context, keyID string) error {
 	return nil
 }
 
+// CreateKeyV2 tries to create a new master key with the given keyID
+// at the KMS connected to a MinIO server.
+func (adm *AdminClient) CreateKeyV2(ctx context.Context, keyID string) error {
```
harshavardhana, line 155:
```suggestion
func (adm *AdminClient) CreateKMSKey(ctx context.Context, keyID string) error {
```

kms-commands.go
```diff
@@ -204,6 +241,26 @@ func (adm *AdminClient) GetKeyStatus(ctx context.Context, keyID string) (*KMSKey
 	return &keyInfo, nil
 }
 
+// GetKeyStatusV2 requests status information about the key referenced by keyID
+// from the KMS connected to a MinIO by performing a Admin-API request.
+// It basically hits the `/minio/admin/v3/kms/key/status` API endpoint.
+func (adm *AdminClient) GetKeyStatusV2(ctx context.Context, keyID string) (*KMSKeyStatus, error) {
```
harshavardhana, line 247:
```suggestion
func (adm *AdminClient) GetKMSKeyStatus(ctx context.Context, keyID string) (*KMSKeyStatus, error) {
```

kms-commands.go
```diff
@@ -107,8 +108,28 @@ func (adm *AdminClient) KMSStatus(ctx context.Context) (KMSStatus, error) {
 	return status, nil
 }
 
+// KMSStatusV2 returns status information about the KMS connected
+// to the MinIO server, if configured.
+func (adm *AdminClient) KMSStatusV2(ctx context.Context) (KMSStatus, error) {
```
harshavardhana, line 113: This can be changed to KMStatus() there is no reason to keep older API? since both return same values no client will break.

harshavardhana (changes requested):

kms-commands.go
harshavardhana: thoughts @aead ^^?

reivaj05: I believe it should be fine as long as [this PR](https://github.com/minio/minio/pull/15737) gets merged before this one

harshavardhana: That's not a problem @reivaj05 madmin-go will have a release tag only meant to be used with latest MinIO release, we can also make madmin-go move to `v2.0.0` so this way it indicates a large breakage.

harshavardhana: can you make the relevant changes? @reivaj05

reivaj05: Sure, will do @harshavardhana

harshavardhana (changes requested):

kms-commands.go
```diff
@@ -41,8 +41,8 @@ type KMSKeyInfo struct {
 
 // KMSPolicyInfo contains policy metadata
 type KMSPolicyInfo struct {
-	CreatedAt string `json:"createdAt"`
-	CreatedBy string `json:"createdBy"`
+	CreatedAt string `json:"created_at"`
+	CreatedBy string `json:"created_by"`
```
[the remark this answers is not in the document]
reivaj05: @harshavardhana This format is coming from KES in (By mistake i handled it as camelcase in my previous PR):

`// PolicyInfo describes a KES policy.
type PolicyInfo struct {
	Name      string    `json:"name"`                 // Name of the policy
	CreatedAt time.Time `json:"created_at,omitempty"` // Point in time when the policy was created
	CreatedBy Identity  `json:"created_by,omitempty"` // Identity that created the policy
}`

but i can add a new struct as wrapper in minio to handle everything in camelcase, what do you think?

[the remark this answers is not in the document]
harshavardhana: oh okay then its fine @reivaj05

harshavardhana (approved):

PR #134, opened 2022-10-10
Add missing info fields

klauspost:
Field are missing to fill ServerInfo.

### DIFF
--- a/info-commands.go
+++ b/info-commands.go
@@ -382,7 +382,7 @@ type ServerProperties struct {
 	GOMAXPROCS     int               `json:"gomaxprocs,omitempty"`
 	NumCPU         int               `json:"num_cpu,omitempty"`
 	RuntimeVersion string            `json:"runtime_version,omitempty"`
-	GCStats        GCStats           `json:"gc_stats,omitempty"`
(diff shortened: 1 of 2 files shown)


info-commands.go
```diff
@@ -369,16 +369,21 @@ type ErasureBackend struct {
...
+	Endpoint       string            `json:"endpoint,omitempty"`
+	Scheme         string            `json:"scheme,omitempty"`
+	Uptime         int64             `json:"uptime,omitempty"`
+	Version        string            `json:"version,omitempty"`
+	CommitID       string            `json:"commitID,omitempty"`
+	Network        map[string]string `json:"network,omitempty"`
+	Disks          []Disk            `json:"drives,omitempty"`
+	PoolNumber     int               `json:"poolNumber,omitempty"`
+	MemStats       MemStats          `json:"mem_stats"`
+	GOMAXPROCS     int               `json:"gomaxprocs,omitempty"`
```
vadmeste, line 382: Maybe GoMaxProcs

klauspost: Yea, I'll change the [existing](https://github.com/minio/madmin-go/blob/d90007c2e5453efaae7862a96dcd3f5fa21981d8/health.go#L801) as well.

PR #146, opened 2022-10-21
Add admin APIs for managing policies

taran-p:
Add new commands to manage policies on users and groups


policy-commands.go
```diff
@@ -225,3 +225,164 @@ func (adm *AdminClient) SetPolicy(ctx context.Context, policyName, entityName st
...
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return p, err
+	}
+
+	if err = json.Unmarshal(b, &p); err != nil {
+		return p, err
+	}
+
+	return p, nil
```
klauspost, line 308: Avoids using temporary buffer and doesn't use old ioutil as bonus.

```suggestion
	dec := json.NewDecoder(resp.Body)
	err = dec.Decode(&p)
	return p, err
```

policy-commands.go
```diff
@@ -225,3 +225,164 @@ func (adm *AdminClient) SetPolicy(ctx context.Context, policyName, entityName st
...
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = json.Unmarshal(data, &p); err != nil {
+		return nil, err
+	}
+
+	return p, nil
```
klauspost, line 387:
```suggestion
	dec := json.NewDecoder(resp.Body)
	err = dec.Decode(&p)
	return p, err
```

klauspost (approved):

PR #148, opened 2022-11-02
Add time zone to time-info

anjalshireesh:
So that check(s) related to time zone can be added in `health-analyzer`


vadmeste (approved):
LGTM

prakashsvmx (approved):

PR #152, opened 2022-11-10
client to server bandwidth measurement API

harshavardhana:


harshavardhana:
PTAL @vadmeste @klauspost

harshavardhana:
NOTE: the measurement here is only for throughput so there is no RX/TX involved in this.

klauspost:
Shouldn't we add the download benchmark right away? Do we have a speedtest endpoint we can pull data from?

harshavardhana:
> Shouldn't we add the download benchmark right away? Do we have a speedtest endpoint we can pull data from?
> 
> In many cases bandwidth will be async.

There is no reason to do that you can saturate the link by simply sending data @klauspost - since its pure network doesn't need write/read like differentiation IMO.

klauspost:
What do you mean? There are many async connection types out there. Measuring upload only gives half the picture.

harshavardhana:
> What do you mean? There are many async connection types out there. Measuring upload only gives half the picture.

We are only validating what is the network throughput from client to server, this doesn't measure anything else. Of course, the server is not sending data back to the client here, but that is not useful IMO. 

Because all we care to show is the X bandwidth fully saturated value what is possible from client to server. In a full duplex connection such as 100g the 12GiB/sec value itself shows what is the overall bandwidth.

harshavardhana:
any other questions @klauspost @vadmeste ?

perf-net.go
```diff
@@ -56,3 +61,81 @@ func (adm *AdminClient) Netperf(ctx context.Context, duration time.Duration) (re
 	err = json.NewDecoder(resp.Body).Decode(&result)
 	return result, err
 }
+
+type NetperfClientResult struct {
+	Endpoint  string `json:"endpoint"`
+	Bandwidth uint64 `json:"bw"`
```
klauspost, line 67:
```suggestion
	UploadBandwidth uint64 `json:"up_bps"`
```

klauspost (changes requested):
> In a full duplex connection such as 100g, the 12GiB/sec value itself shows what is the overall bandwidth.

There are many async connections that do not provide close to the same upload/download speeds.

I don't mind that you don't want to add download test now. But then let's name the field what it is, so we can add download later.

harshavardhana:
Will re-submit when we need this..

PR #153, opened 2022-11-13
Add a humanized time elapsed field to LockEntry

krisis:
This would allow consumers of `mc support top locks --json` subcommand to know the time for which the lock was held at the time of receving the json output.

### DIFF
--- a/top-commands.go
+++ b/top-commands.go
@@ -31,13 +31,14 @@ import (
 // servers holding the lock, source on the client machine,
 // ID, type(read or write) and time stamp.
 type LockEntry struct {
-	Timestamp  time.Time `json:"time"`       // When the lock was first granted
-	Resource   string    `json:"resource"`   // Resource contains info like bucket+object
-	Type       string    `json:"type"`       // Type indicates if 'Write' or 'Read' lock
-	Source     string    `json:"source"`     // Source at which lock was granted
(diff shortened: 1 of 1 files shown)


top-commands.go
```diff
@@ -32,6 +32,7 @@ import (
 // ID, type(read or write) and time stamp.
 type LockEntry struct {
 	Timestamp  time.Time `json:"time"`       // When the lock was first granted
+	Elapsed    string    `json:"elapsed"`    // Humanized duration for which lock has been held
```
harshavardhana, line 35: Can this not be done on the `mc` side? such as `time.Since(lk.Timestamp)` ?  - we can then humanize it. Adding a string representation from the server is not necessary and an overkill for a data structure IMO.

harshavardhana (changes requested):

top-commands.go
krisis: Here are the reasons why it makes sense for the server to compute duration for which locks were held
- top locks API response should be self contained
- while elapsed time can be derived at client side, computing it at the server-side makes it relative to server system time as it should be.

harshavardhana: I am against stringifying in this manner and sending humanized forms in a data structure. That is not what Elapsed is meant for if you want to compute we should keep it as `time.Duration` instead.

harshavardhana (approved):

PR #159, opened 2022-11-29
Fix #157, remove host.Warnings check

### LINKED ISSUE #157: Unable to build with github.com/shirou/gopsutil/v3 v3.22.10
I am working on packaging this library for use in Debian, and have encountered the following build error when using github.com/shirou/gopsutil/v3 v3.22.10 (the current version of that library and recently updated in Debian):

```
# github.com/minio/madmin-go
src/github.com/minio/madmin-go/health.go:328:36: undefined: host.Warnings
```

It appears that commit https://github.com/shirou/gopsutil/commit/dbc0f20fe38841a0c1509de9dba5084b7c2c09d9 moved `host/types.go` -> `internal/common/warnings.go`.

mrusme:
Warnings were moved to gopsutil internal:
https://github.com/shirou/gopsutil/compare/v3.22.9...v3.22.10

### DIFF
--- a/health.go
+++ b/health.go
@@ -323,12 +323,7 @@ func GetOSInfo(ctx context.Context, addr string) OSInfo {
 	}
 	osInfo.Info.KernelVersion = kr
 
-	osInfo.Sensors, err = host.SensorsTemperaturesWithContext(ctx)
-	if err != nil {
-		if _, isWarningErr := err.(*host.Warnings); !isWarningErr {
-			osInfo.Error = err.Error()
-		}
-	}
+	osInfo.Sensors, _ = host.SensorsTemperaturesWithContext(ctx)
 
 	return osInfo
 }


health.go
```diff
@@ -325,9 +325,7 @@ func GetOSInfo(ctx context.Context, addr string) OSInfo {
 
 	osInfo.Sensors, err = host.SensorsTemperaturesWithContext(ctx)
 	if err != nil {
-		if _, isWarningErr := err.(*host.Warnings); !isWarningErr {
-			osInfo.Error = err.Error()
-		}
+		osInfo.Error = err.Error()
```
harshavardhana, line 328: We only need this error if its not host.Warnings() @mrusme

harshavardhana (changes requested):
This is not an expected change.

health.go
mrusme: To get this right, in case the following issue should arise, the error would not be needed?

https://github.com/shirou/gopsutil/blob/34cc43d282f68f93b2f9c5be19d0c4a6beb7d39d/host/host_linux.go#L386

In that case I guess the only option would be to remove the check altogether. What do you think?

harshavardhana: yes that is acceptable @mrusme

harshavardhana (approved):

anjalshireesh (approved):

klauspost:
Seems like there are plans to move it again: https://github.com/shirou/gopsutil/pull/1379

klauspost (approved):

PR #164, opened 2022-12-20
Add bootstrap trace to trace early MinIO cluster events

vadmeste:
Trace early events of MinIO cluster bootstrap

### DIFF
--- a/service-commands.go
+++ b/service-commands.go
@@ -106,6 +106,7 @@ type ServiceTraceOpts struct {
 	BatchReplication  bool
 	Rebalance         bool
 	ReplicationResync bool
+	Bootstrap         bool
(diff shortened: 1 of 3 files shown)


klauspost:
I almost feel we should have a buffer for these events. There is a fairly small window to connect.

harshavardhana (approved):

vadmeste:
> I almost feel we should have a buffer for these events. There is a fairly small window to connect.

Umm, yes the buffer is good when everything is stuck and there is nothing in progress, but it is little difficult for this to happen. We can do this later.

vadmeste:
can we merge this @klauspost ?

klauspost (approved):

PR #165, opened 2022-12-21
Change policy attach/detach API structure

taran-p:
Slightly restructures attach/detach for policies; sends information in PolicyAssociationReq struct instead of plain.


kannappanr (approved):

vadmeste (approved):
LGTM

vadmeste (changes requested):
LGTM but this is a breaking change, we should briefly talk about this with @harshavardhana

harshavardhana (changes requested):
@taran-p please add this as a new API instead of changing existing ones, we can remove this in future when we move to v3

taran-p:
@harshavardhana Can do, but I was wondering if it would make sense to change these APIs since I added them recently and they aren't used for anything yet as far as I am aware

harshavardhana:
> @harshavardhana Can do, but I was wondering if it would make sense to change these APIs since I added them recently and they aren't used for anything yet as far as I am aware

So if it's fairly new then it's not a problem.

harshavardhana (approved):

PR #169, opened 2023-01-14
Simplify merging the heal status of different nodes

vadmeste:
Merge now assumes that each node reports same 
sets but with different disks. This is needed so the 
server can optimize its code to use LocalStorageInfo() 
call for each node instead of StorageInfo() which is 
repetitive and costly.


klauspost (approved):
LGTM

harshavardhana (approved):

PR #170, opened 2023-01-18
Add duration to Add/Update service account requests

Praveenrajmani:
This allows to add/modify the expiration of service accounts


user-commands.go
```diff
@@ -303,6 +303,7 @@ type AddServiceAccountReq struct {
 	AccessKey  string          `json:"accessKey,omitempty"`
 	SecretKey  string          `json:"secretKey,omitempty"`
 	Comment    string          `json:"comment,omitempty"`
+	Duration   time.Duration   `json:"duration,omitempty"`
```
klauspost, line 306: From an API perspective I think `Expire *time.Time` is more clear, and leave the time calculation to the clients.

Praveenrajmani: Ok. For this, the clients need to calculate the time w.r.t the server timezones. The expiration time will be invalidated accordingly if it falls behind the server's time.

^^ @klauspost

klauspost: Time zones shouldn't be a problem. When sending as JSON the time zone is included, so you are sending an absolute time.

The server can validate and convert to UTC.

user-commands.go
```diff
@@ -434,11 +436,12 @@ func (adm *AdminClient) ListServiceAccounts(ctx context.Context, user string) (L
...
-	AccountStatus string `json:"accountStatus"`
-	ImpliedPolicy bool   `json:"impliedPolicy"`
-	Policy        string `json:"policy"`
-	Comment       string `json:"comment"`
+	ParentUser    string    `json:"parentUser"`
+	AccountStatus string    `json:"accountStatus"`
+	ImpliedPolicy bool      `json:"impliedPolicy"`
+	Policy        string    `json:"policy"`
+	Comment       string    `json:"comment"`
+	Expiration    time.Time `json:"expiration,omitempty"`
```
klauspost, line 444: So we don't have to deal with zero values.
```suggestion
	Expiration    *time.Time `json:"expiration,omitempty"`
```

klauspost (approved):
LGTM

harshavardhana (approved):

examples/service-account.go
```diff
@@ -44,8 +45,10 @@ func main() {
 	ctx := context.Background()
 
 	// add service account
+	expiration := time.Now().Add(30 * time.Minute)
 	addReq := madmin.AddServiceAccountReq{
 		TargetUser: "my-username",
+		Expiration: &expiration,
```
kaankabalak, line 51: Hi @Praveenrajmani, when I don't send in the `expiration` field here, the API seems to return an `expiration` value of `1970-01-01T00:00:00Z`. Is this expected behavior?

Screen Shot 2023-05-08 at 7 34 46 PM

You can reproduce the issue by applying the following diff:

```diff
diff --git a/examples/service-account.go b/examples/service-account.go
index 73856cf..88b02a9 100644
--- a/examples/service-account.go
+++ b/examples/service-account.go
@@ -45,10 +45,8 @@ func main() {
 	ctx := context.Background()
 
 	// add service account
-	expiration := time.Now().Add(30 * time.Minute)
 	addReq := madmin.AddServiceAccountReq{
 		TargetUser: "my-username",
-		Expiration: &expiration,
 	}
 	addRes, err := madminClient.AddServiceAccount(context.Background(), addReq)
 	if err != nil {
```

harshavardhana: Well that is because it is considered no expiration @kaankabalak

PR #171, opened 2023-01-31
add Transport in api.Options

sebastiaopamplona:
[minio-go](https://github.com/minio/minio-go) supports the option Transport (see [here](https://github.com/minio/minio-go/blob/f5f1604f764e0543943961efefac0e6dc56ee4b2/api.go#L105)).

It would be nice if madmin-go also supported it :)

### DIFF
--- a/api.go
+++ b/api.go
@@ -95,8 +95,9 @@ const (
 
 // Options for New method
 type Options struct {
-	Creds  *credentials.Credentials
-	Secure bool
+	Creds     *credentials.Credentials
+	Secure    bool
+	Transport http.RoundTripper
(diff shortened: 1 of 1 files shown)


klauspost (approved):
LGTM

harshavardhana (changes requested):
Can you document which one takes precedence this one or SetCustomTransport() ?

Also add a comment to deprecate SetCustomTransport

klauspost:
Is this really needed then?

sebastiaopamplona:
> Is this really needed then?

You're right, it's not. Sorry for the confusion!

PR #172, opened 2023-02-01
Return malicious data error with unexpected stream header

vadmeste:
When decrypting data with a secret key, that is used in decrypting
 admin requests or old MinIO backend configuration, consider an 
incomplete header as malicious data instead of returning io.EOF. 
That way, the error will be more understood by the end user.

### DIFF
--- a/encrypt.go
+++ b/encrypt.go
@@ -113,7 +113,7 @@ var ErrMaliciousData = sio.NotAuthentic
 // EncryptData. Otherwise, the decryption will fail.
 func DecryptData(password string, data io.Reader) ([]byte, error) {
 	// Parse the stream header
-	var hdr [41]byte
+	var hdr [32 + 1 + 8]byte
 	if _, err := io.ReadFull(data, hdr[:]); err != nil {
 		if err == io.EOF {
 			// Incomplete header, return malicious data
--- a/encrypt.go
+++ b/encrypt.go
@@ -104,6 +104,9 @@ func EncryptData(password string, data []byte) ([]byte, error) {
 // decrypted by provided credentials.
 var ErrMaliciousData = sio.NotAuthentic
 
+// ErrUnexpectedHeader indicates that the data stream returned unexpected header
+var ErrUnexpectedHeader = errors.New("unexpected header")
+
 // DecryptData decrypts the data with the key derived
 // from the salt (part of data) and the password using
 // the PBKDF used in EncryptData. DecryptData returns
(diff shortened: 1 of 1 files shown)


encrypt.go
```diff
@@ -113,35 +112,30 @@ var ErrMaliciousData = sio.NotAuthentic
...
-	if _, err := io.ReadFull(data, id[:]); err != nil {
-		return nil, err
-	}
-	if _, err := io.ReadFull(data, nonce[:]); err != nil {
+	// Parse the stream header
+	var hdr [41]byte
+	if _, err := io.ReadFull(data, hdr[:]); err != nil {
+		if err == io.EOF {
+			// Incomplete header, return malicious data
+			return nil, ErrMaliciousData
```
klauspost, line 120:
```suggestion
			return nil, io.ErrUnexpectedEOF
```

encrypt.go
```diff
@@ -113,35 +112,30 @@ var ErrMaliciousData = sio.NotAuthentic
...
-
-	if _, err := io.ReadFull(data, salt[:]); err != nil {
-		return nil, err
-	}
-	if _, err := io.ReadFull(data, id[:]); err != nil {
-		return nil, err
-	}
-	if _, err := io.ReadFull(data, nonce[:]); err != nil {
+	// Parse the stream header
+	var hdr [41]byte
```
klauspost, line 116:
```suggestion
	var hdr [32+1+8]byte
```

klauspost (changes requested):
Wouldn't call it "malicious", just unexpected.

encrypt.go
vadmeste: @klauspost I think malicious data is better here. The point is to return a more meangniful error to the user to understand what is going on.

Besides, malicious data should be acceptable as well when we encounter io.EOF specifically. When there is io.EOF, it means the stream is ended in an expected way, and if we did not receive enough data for the header, I think we can consider it

aead (approved):

harshavardhana (approved):

vadmeste:
@klauspost can you take a look again ?

encrypt.go
harshavardhana: Even io.ErrUnexpectedEOF is okay - we don't really ignore it anyways for this caller. We can treat that at the caller layer on what the meaning of this io.ErrUnexpectedEOF must be.

encrypt.go
```diff
@@ -113,35 +112,30 @@ var ErrMaliciousData = sio.NotAuthentic
...
-		return nil, err
-	}
-	if _, err := io.ReadFull(data, id[:]); err != nil {
-		return nil, err
-	}
-	if _, err := io.ReadFull(data, nonce[:]); err != nil {
+	// Parse the stream header
+	var hdr [32 + 1 + 8]byte
+	if _, err := io.ReadFull(data, hdr[:]); err != nil {
+		if err == io.EOF {
```
harshavardhana, line 118: let's use `errors.Is(err, io.EOF)`

harshavardhana (changes requested):

vadmeste:
Made more changes, let me know what do you think @klauspost and @harshavardhana

harshavardhana (approved):

PR #173, opened 2023-02-09
Add per erasure set info in server info

vadmeste:
Add per erasure set information, that will be shown in mc admin info 
output or AdminClient.ServerInfo().

### DIFF
--- a/info-commands.go
+++ b/info-commands.go
@@ -222,6 +222,17 @@ func (adm *AdminClient) DataUsageInfo(ctx context.Context) (DataUsageInfo, error
 	return dataUsageInfo, nil
 }
 
+// ErasureSetInfo provides information per erasure set
+type ErasureSetInfo struct {
+	ID            int    `json:"id"`
+	RawUsage      uint64 `json:"rawUsage"`
+	RawCapacity   uint64 `json:"rawCapacity"`
+	Usage         uint64 `json:"usage"`
+	ObjectsCount  uint64 `json:"objectsCount"`
+	VersionsCount uint64 `json:"versionsCount"`
+	HealDisks     int    `json:"healDisks"`
+}
+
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

info-commands.go
```diff
@@ -236,6 +247,8 @@ type InfoMessage struct {
 	Services     Services           `json:"services,omitempty"`
 	Backend      interface{}        `json:"backend,omitempty"`
 	Servers      []ServerProperties `json:"servers,omitempty"`
+
+	Pools map[int]map[int]ErasureSetInfo `json:"pools,omitempty"`
```
klauspost, line 251: `map[int]` cannot be used in JSON. It must be `map[string]` or arrays.

harshavardhana: doesn't go automatically convert this to string? @klauspost

klauspost (changes requested):

vadmeste:
@klauspost it works fine for me. You just need to test https://github.com/minio/minio/pull/16583

vadmeste:
Oh, also you need to edit go.mod in github.com/minio/mc

```
replace github.com/minio/madmin-go/v2 => github.com/vadmeste/madmin-go/v2 v2.0.7-0.20230209161037-4883f168c2c6
```

then `mc admin info --json <alias> | jq .info.pools`

info-commands.go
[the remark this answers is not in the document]
klauspost: Oh wow, they added that in Go 1.7

> In earlier versions of Go, this package only supported encoding and decoding maps using keys with string types. Go 1.7 adds support for maps using keys with integer types: the encoding uses a quoted decimal representation as the JSON key.

Seems risky to use such a recent feature :D

klauspost (approved):
LGTM

PR #174, opened 2023-02-14
speedtest: Add 'noclear' flag to keep the generated data

vadmeste:
For mc support perf <alias>, the default behavior is to remove 
the generated objects when doing object speed test. Adding
 a flag to offer a possibility to keep the generated data, so we 
can have a way to quickly generate data of any size.

e.g.
mc support perf object <alias> --duration 30m --keep-data --size 1K

### DIFF
--- a/perf-object.go
+++ b/perf-object.go
@@ -65,6 +65,7 @@ type SpeedtestOpts struct {
 	Autotune     bool          // Enable autotuning
 	StorageClass string        // Choose type of storage-class to be used while performing I/O
 	Bucket       string        // Choose a custom bucket name while performing I/O
+	KeepData     bool          // Avoid cleanup after running an object speed test
 }
 
 // Speedtest - perform speedtest on the MinIO servers
@@ -96,6 +97,9 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (chan
 	}
 	if opts.Autotune {
 		queryVals.Set("autotune", "true")
+	}
+	if opts.KeepData {
+		queryVals.Set("keep-data", "true")
 	}
 	resp, err := adm.executeMethod(ctx,
 		http.MethodPost, requestData{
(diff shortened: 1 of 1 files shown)


perf-object.go
```diff
@@ -65,6 +65,7 @@ type SpeedtestOpts struct {
 	Autotune     bool          // Enable autotuning
 	StorageClass string        // Choose type of storage-class to be used while performing I/O
 	Bucket       string        // Choose a custom bucket name while performing I/O
+	KeepData     bool          // Avoid cleanup after running an object speed test
```
harshavardhana, line 68:
```suggestion
	NoClear     bool          // Avoid cleanup after running an object speed test
```

perf-object.go
```diff
@@ -97,6 +98,9 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (chan
 	if opts.Autotune {
 		queryVals.Set("autotune", "true")
 	}
+	if opts.KeepData {
```
harshavardhana, line 101:
```suggestion
	if opts.NoClear {
```

perf-object.go
```diff
@@ -97,6 +98,9 @@ func (adm *AdminClient) Speedtest(ctx context.Context, opts SpeedtestOpts) (chan
...
+		queryVals.Set("keep-data", "true")
```
harshavardhana, line 102:
```suggestion
		queryVals.Set("noclear", "true")
```

harshavardhana (changes requested):

vadmeste:
@harshavardhana it looks like this PR is overkill since a single 'warp' cli command is enough, what do you think?

vadmeste:
We just this for this PR https://github.com/actions/go-versions/pull/70 to be merged to fix vulncheck complaints

vadmeste:
@harshavardhana can you take a look ?

harshavardhana (approved):

PR #176, opened 2023-02-20
Add custom key/val to TraceInfo

klauspost:
Add a custom key+val field for adding additional info with some structure.


vadmeste (approved):
LGTM

harshavardhana (approved):

PR #182, opened 2023-03-22
add field to PeerInfo struct

poornas:
for passing sync mode in site replication

### DIFF
--- a/batch-job.go
+++ b/batch-job.go
@@ -205,7 +205,7 @@ type GenerateBatchJobOpts struct {
 
 // GenerateBatchJob creates a new job template from standard template
 // TODO: allow configuring yaml values
-func (adm *AdminClient) GenerateBatchJob(ctx context.Context, opts GenerateBatchJobOpts) (string, error) {
+func (adm *AdminClient) GenerateBatchJob(_ context.Context, opts GenerateBatchJobOpts) (string, error) {
 	switch opts.Type {
 	case BatchJobReplicate:
 		// TODO: allow configuring the template to fill values from GenerateBatchJobOpts
--- a/cluster-commands.go
+++ b/cluster-commands.go
@@ -151,7 +151,18 @@ type PeerInfo struct {
 	Name     string `json:"name"`
 	// Deployment ID is useful as it is immutable - though endpoint may
 	// change.
-	DeploymentID string `json:"deploymentID"`
+	DeploymentID string     `json:"deploymentID"`
+	SyncState    SyncStatus `json:"sync"` // whether to enable| disable synchronous replication
+}
+
+type SyncStatus string // change in sync state
+const (
+	SyncEnabled  SyncStatus = "enable"
+	SyncDisabled SyncStatus = "disable"
+)
+
+func (s SyncStatus) Empty() bool {
+	return s != SyncDisabled && s != SyncEnabled
 }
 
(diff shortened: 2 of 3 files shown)


batch-job.go
```diff
@@ -205,7 +205,7 @@ type GenerateBatchJobOpts struct {
 
 // GenerateBatchJob creates a new job template from standard template
 // TODO: allow configuring yaml values
-func (adm *AdminClient) GenerateBatchJob(ctx context.Context, opts GenerateBatchJobOpts) (string, error) {
+func (adm *AdminClient) GenerateBatchJob(opts GenerateBatchJobOpts) (string, error) {
```
harshavardhana, line 208: unrelated change please revert - pass `_ context.Context` like this, we need this for future purposes.

cluster-commands.go
```diff
@@ -151,7 +151,18 @@ type PeerInfo struct {
 	Name     string `json:"name"`
 	// Deployment ID is useful as it is immutable - though endpoint may
 	// change.
-	DeploymentID string `json:"deploymentID"`
+	DeploymentID string     `json:"deploymentID"`
+	SyncState    SyncStatus `json:"sync"` // whether to enable| disable synchronous replication
```
harshavardhana, line 155: why is this not a boolean? @poornas

health.go
```diff
@@ -330,7 +330,7 @@ func GetOSInfo(ctx context.Context, addr string) OSInfo {
 
 // GetSysConfig returns config values from the system
 // (only those affecting minio performance)
-func GetSysConfig(ctx context.Context, addr string) SysConfig {
+func GetSysConfig(addr string) SysConfig {
```
harshavardhana, line 333: unrelated change please revert

health.go
```diff
@@ -356,7 +356,7 @@ func GetSysConfig(ctx context.Context, addr string) SysConfig {
 }
 
 // GetSysServices returns info of sys services that affect minio
-func GetSysServices(ctx context.Context, addr string) SysServices {
+func GetSysServices(addr string) SysServices {
```
harshavardhana, line 359: unrelated change please revert

health.go
```diff
@@ -397,7 +397,7 @@ func getSELinuxInfo() (SysService, error) {
 }
 
 // GetSysErrors returns issues in system setup/config
-func GetSysErrors(ctx context.Context, addr string) SysErrors {
+func GetSysErrors(addr string) SysErrors {
```
harshavardhana, line 400: unrelated change please revert

harshavardhana (changes requested):

cluster-commands.go
harshavardhana: IMO we can make `enable/disable` an `mc` level construct instead of having this via SDK. It is very simple to just have `SyncReplication bool`

poornas: Passing a bool would also entail an additional flag to know if isSyncReplicationSet - otherwise an endpoint change can overwrite the sync mode setting on the server

health.go
poornas: to address lint failures

cluster-commands.go
harshavardhana: Understood.

harshavardhana (approved):

PR #190, opened 2023-04-28
info: Deprecate the Backend interface type

vadmeste:
Gateway and FS are removed, so Backend with interface type is deprecated. 
A new field called ErasureInfo will be added but the code will still support
 the old json format as well when there is no valid information for 
ErasureInfo structure.


klauspost:
I thought/hoped you also removed `type ObjectLayer interface{...}`, now that there is only one implementation. I've begun (twice), but never found the time to finish it ;)

info-commands.go
```diff
@@ -246,7 +246,8 @@ type InfoMessage struct {
 	Versions     Versions           `json:"versions,omitempty"`
 	Usage        Usage              `json:"usage,omitempty"`
 	Services     Services           `json:"services,omitempty"`
-	Backend      interface{}        `json:"backend,omitempty"`
+	Backend      interface{}        `json:"backend,omitempty"` // Deprecated Apr 2023
+	ErasureInfo  ErasureBackend     `json:"erasureInfo,omitempty"`
```
klauspost, line 250: It seems like a hard-to-use deprecation.

Could you make it so `ErasureInfo` is always populated - even if an older server sends it as `Backend`?

Maybe instead of two fields do:
```suggestion
	Backend      ErasureBackend     `json:"backend"`
```
... and do a custom UnmarshalJSON on ErasureBackend, that handles the old types gracefully. That would simplify the deprecation and be backwards compatible both with "new mc -> old minio" and "old mc -> new minio".

(Also omitempty is pointless on structs)

klauspost (changes requested):

info-commands.go
klauspost: Actually you even need a custom UnmarshalJSON, since FSBackend will just be ErasureBackend with only "Type" set.

vadmeste:
@klauspost the point of this PR is to remove old code and to simplify mc code as well.

I also found out that it is possible to marshal ErasureBackend{} structure with a FS or gateway server, so I simplified the code even further. Please take a look again.

klauspost (approved):
Exactly :D

harshavardhana (approved):

PR #193, opened 2023-05-09
replication: Add MRF backlog API

poornas:
Return most recent failed replication entries


examples/replicate-mrf.go
```diff
@@ -0,0 +1,50 @@
+//go:build ignore
+// +build ignore
+
+// Copyright (c) 2015-2022 MinIO, Inc.
```
harshavardhana, line 4:
```suggestion
// Copyright (c) 2015-2023 MinIO, Inc.
```

harshavardhana (changes requested):

replication-api.go
```diff
@@ -109,3 +109,61 @@ func (adm *AdminClient) BucketReplicationDiff(ctx context.Context, bucketName st
 	// Returns the diff channel, for caller to start reading from.
 	return diffCh
 }
+
+// ReplicationMRF rrepresents MRF backlog for a bucket
```
klauspost, line 113:
```suggestion
// ReplicationMRF represents MRF backlog for a bucket
```

replication-api.go
```diff
@@ -109,3 +109,61 @@ func (adm *AdminClient) BucketReplicationDiff(ctx context.Context, bucketName st
...
+		if node != "" {
+			queryValues.Set("node", node)
+		}
+		reqData := requestData{
+			relPath:     adminAPIPrefix + "/replication/mrf",
+			queryValues: queryValues,
+		}
+
+		// Execute PUT on /minio/admin/v3/mrf to get mrf backlog for a bucket.
+		resp, err := adm.executeMethod(ctx, http.MethodPost, reqData)
```
klauspost, line 142: Put or Post?

Really it should be a GET ;)

klauspost (commented):
Is this an event stream of failures?

I was initially expecting something that would fit into the Metrics API, but looking more at it, it does seem different.

Are you adding a Metrics type later that aggregates replication stats?

poornas:
> Is this an event stream of failures?
> 
> I was initially expecting something that would fit into the Metrics API, but looking more at it, it does seem different.
> 
> Are you adding a Metrics type later that aggregates replication stats?

this is an event stream of failures, will check out the Metrics API in other PR that does the replication stats.

api-log.go
```diff
@@ -102,7 +102,7 @@ func (adm AdminClient) GetLogs(ctx context.Context, node string, lineCnt int, lo
 				queryValues: urlValues,
 			}
 			// Execute GET to call log handler
-			resp, err := adm.executeMethod(ctx, http.MethodGet, reqData)
+			resp, err := adm.executeMethod(ctx, http.MethodPost, reqData)
```
klauspost, line 105: ? Seems like an unintentional change?

replication-api.go
```diff
@@ -109,3 +117,60 @@ func (adm *AdminClient) BucketReplicationDiff(ctx context.Context, bucketName st
 	// Returns the diff channel, for caller to start reading from.
 	return diffCh
 }
+
+// ReplicationMRF represents MRF backlog for a bucket
+type ReplicationMRF struct {
+	NodeName   string `json:"nodeName" msg:"n"`
```
vadmeste, line 123: Shouldn't we add a Version here, so it is easy to error out when there is an incompatibility in the format and we can make madmin-go supports multiple versions. Maybe we can use major/minor version as well ?

vadmeste (changes requested):
A comment

api-log.go
poornas: thanks, removed this

replication-api.go
poornas: done

poornas:
@vadmeste , @harshavardhana PTAL

vadmeste (approved):
LGTM

vadmeste (approved):
LGTM

harshavardhana (approved):

PR #197, opened 2023-06-01
Added endpoint and version

shtripat:
This PR adds `endpoint` and `version` attributes to a KMS instance. Also for `admin info` command, `services.kms` would now be a list of KMS instances.

### DIFF
--- a/info-commands.go
+++ b/info-commands.go
@@ -275,7 +275,8 @@ func (info InfoMessage) StandardParity() int {
 
 // Services contains different services information
 type Services struct {
-	KMS           []KMS                         `json:"kms,omitempty"`
+	KMS           KMS                           `json:"kms,omitempty"` // deprecated july 2023
+	KMSV2         []KMS                         `json:"kmsv2,omitempty"`
 	LDAP          LDAP                          `json:"ldap,omitempty"`
 	Logger        []Logger                      `json:"logger,omitempty"`
 	Audit         []Audit                       `json:"audit,omitempty"`
--- a/info-commands.go
+++ b/info-commands.go
@@ -276,7 +276,7 @@ func (info InfoMessage) StandardParity() int {
 // Services contains different services information
 type Services struct {
 	KMS           KMS                           `json:"kms,omitempty"` // deprecated july 2023
-	KMSV2         []KMS                         `json:"kmsv2,omitempty"`
+	KMSStatus     []KMS                         `json:"kmsStatus,omitempty"`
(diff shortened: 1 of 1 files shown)


info-commands.go
```diff
@@ -316,9 +316,11 @@ type TierStats struct {
 
 // KMS contains KMS status information
 type KMS struct {
-	Status  string `json:"status,omitempty"`
-	Encrypt string `json:"encrypt,omitempty"`
-	Decrypt string `json:"decrypt,omitempty"`
+	Status   string `json:"status,omitempty"`
+	Encrypt  string `json:"encrypt,omitempty"`
+	Decrypt  string `json:"decrypt,omitempty"`
```
vadmeste, line 321: Will those fields contain an error message if it fails to encrypt/decrypt ?

shtripat: @vadmeste yes, that logic remains as old one. Just that all this gets reported at each instance level.
So you feel we should have separate error fields for reporting the encrypt/decrypt errors and make these as booleans?

klauspost (approved):

info-commands.go
```diff
@@ -275,7 +275,7 @@ func (info InfoMessage) StandardParity() int {
 
 // Services contains different services information
 type Services struct {
-	KMS           KMS                           `json:"kms,omitempty"`
+	KMS           []KMS                         `json:"kms,omitempty"`
```
vadmeste, line 278: This can break the compabitility of newer mc with an older MinIO server, can we add a new fields instead and add comment for KMS field (Deprecated July 2023)

vadmeste (commented):
One comment

info-commands.go
shtripat: Sure. Thanks for suggestions. Will update it.

info-commands.go
```diff
@@ -275,7 +275,8 @@ func (info InfoMessage) StandardParity() int {
 
 // Services contains different services information
 type Services struct {
-	KMS           KMS                           `json:"kms,omitempty"`
+	KMS           KMS                           `json:"kms,omitempty"` // deprecated july 2023
+	KMSV2         []KMS                         `json:"kmsv2,omitempty"`
```
harshavardhana, line 279:
```suggestion
	KMSStatus         []KMS                         `json:"kmsStatus,omitempty"`
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #198, opened 2023-06-14
Add response type to user/group policy association

donatello:
This should bump up to v3.0.0 as it is backwards incompatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ package main
 import (
     "fmt"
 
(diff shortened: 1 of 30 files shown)


harshavardhana:
Can you update `go.mod` to indicate the v3 change?

donatello:
@harshavardhana done.

harshavardhana (approved):

PR #200, opened 2023-06-20
support perf site replication

jiuker:
support perf site replication

### DIFF
--- a/perf-site-replication.go
+++ b/perf-site-replication.go
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2015-2022 MinIO, Inc.
+// Copyright (c) 2015-2023 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
--- a/perf-site-replication.go
+++ b/perf-site-replication.go
@@ -32,7 +32,7 @@ type SiteNetPerfNodeResult struct {
 	Endpoint  string `json:"endpoint"`
 	TX        uint64 `json:"tx"`
 	TXTotalDuration   time.Duration `json:"txTotalDuration"`
-	RX        uint64 `json:"rx"`
+	RX        uint64 `json:"rx"` // received rate in bytes 
 	RXTotalDuration   time.Duration `json:"rxTotalDuration"`
 	TotalConn uint64 `json:"totalConn"`
 	Error     string `json:"error,omitempty"`
--- a/perf-site-replication.go
+++ b/perf-site-replication.go
@@ -33,7 +33,7 @@ type SiteNetPerfNodeResult struct {
 	TX        uint64 `json:"tx"`
 	TXTotalDuration   time.Duration `json:"txTotalDuration"`
 	RX        uint64 `json:"rx"`
-	RxDurMs   uint64 `json:"rxDurMs"`
+	RXTotalDuration   time.Duration `json:"rxTotalDuration"`
 	TotalConn uint64 `json:"totalConn"`
 	Error     string `json:"error,omitempty"`
 }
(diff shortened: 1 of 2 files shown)


perf-site-replication.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// SiteReplicationperfNodeResult - stats from each server
+type SiteReplicationperfNodeResult struct {
```
poornas, line 31:
```suggestion
type SiteReplicationPerfNodeResult struct {
```

perf-site-replication.go
```diff
@@ -0,0 +1,62 @@
...
+	Endpoint string `json:"endpoint"`
+	TX       uint64 `json:"tx"`
+	RX       uint64 `json:"rx"`
+	Error    string `json:"error,omitempty"`
+}
+
+// SiteReplicationperfResult - aggregate results from all servers
+type SiteReplicationperfResult struct {
```
poornas, line 39:
```suggestion
type SiteReplicationPerfResult struct {
```

perf-site-replication.go
```diff
@@ -0,0 +1,62 @@
...
+	"net/url"
+	"time"
+)
+
+// SiteNetPerfNodeResult  - stats from each server
+type SiteNetPerfNodeResult struct {
+	Endpoint string `json:"endpoint"`
+	TX       uint64 `json:"tx"`
+	RX       uint64 `json:"rx"`
+	Error    string `json:"error,omitempty"`
```
klauspost, line 35: Don't we need a time indication here, so we know the time frame spent transferring RX/TX?

I feel that should be added for completeness, so we don't have to rely on the query to parse the result.

klauspost: Don't we measure ping latency? Or is that out of scope?

klauspost: Also add the number of concurrent connections per peer.

perf-site-replication.go
```diff
@@ -0,0 +1,65 @@
...
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// SiteNetPerfNodeResult  - stats from each server
+type SiteNetPerfNodeResult struct {
+	Endpoint  string `json:"endpoint"`
+	TX        uint64 `json:"tx"`
+	TxDurMs   uint64 `json:"txDurMs"`
```
harshavardhana, line 34:
```suggestion
	TXTotalDuration   time.Duration `json:"txTotalDuration"`
```

perf-site-replication.go
```diff
@@ -0,0 +1,65 @@
...
+	RX        uint64 `json:"rx"`
+	RxDurMs   uint64 `json:"rxDurMs"`
```
harshavardhana, line 36:
```suggestion
	RXTotalDuration   time.Duration `json:"rxTotalDuration"`
```

perf-site-replication.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// SiteNetPerfNodeResult  - stats from each server
+type SiteNetPerfNodeResult struct {
+	Endpoint  string `json:"endpoint"`
+	TX        uint64 `json:"tx"`
```
harshavardhana, line 33:
```suggestion
	TX        uint64 `json:"tx"` // transfer rate in bytes
```

perf-site-replication.go
```diff
@@ -0,0 +1,65 @@
...
+	TxDurMs   uint64 `json:"txDurMs"`
+	RX        uint64 `json:"rx"`
```
harshavardhana, line 35:
```suggestion
	RX        uint64 `json:"rx"` // received rate in bytes 
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #201, opened 2023-06-21
Add user auth info to UserInfo struct

donatello:


user-commands.go
```diff
@@ -124,8 +124,16 @@ const (
 	AccountDisabled AccountStatus = "disabled"
 )
 
+// UserAuthInfo contains info about how the user is authenticated.
+type UserAuthInfo struct {
+	Type             string `json:"type"`
+	AuthServer       string `json:"authServer,omitempty"`
+	AuthServerUserID string `json:"authServerUserID,omitempty"`
```
harshavardhana, line 131: what are the same values here? and why is it string as in free form?

harshavardhana (changes requested):

user-commands.go
```diff
@@ -124,8 +124,24 @@ const (
 	AccountDisabled AccountStatus = "disabled"
 )
 
+// UserAuthInfo contains info about how the user is authenticated.
+type UserAuthInfo struct {
+	// Type is ldap / builtin / oidc, etc
+	Type string `json:"type"`
```
harshavardhana, line 130:
```suggestion
	Type UserAuthType `json:"type"`
```

harshavardhana (changes requested):

user-commands.go
donatello: Yep, updated now.

poornas (approved):

harshavardhana (approved):

PR #208, opened 2023-07-08
Support mc top net

jiuker:
Support mc top net


metrics.go
```diff
@@ -135,6 +138,7 @@ type RealtimeMetrics struct {
 	Aggregated Metrics               `json:"aggregated"`
 	ByHost     map[string]Metrics    `json:"by_host,omitempty"`
 	ByDisk     map[string]DiskMetric `json:"by_disk,omitempty"`
+	ByNet      map[string]NetMetrics `json:"by_net,omitempty"`
```
klauspost, line 141: Do you mean "By NIC"? Not really sure what "net" refers to here.

What are they added by?

metrics.go
```diff
@@ -206,6 +210,14 @@ func (r *RealtimeMetrics) Merge(other *RealtimeMetrics) {
 	for disk, metrics := range other.ByDisk {
 		r.ByDisk[disk] = metrics
 	}
+
+	// merge NetStats
+	if r.ByNet == nil && len(other.ByNet) > 0 {
+		r.ByNet = make(map[string]NetMetrics, len(other.ByNet))
+	}
+	for disk, metrics := range other.ByNet {
```
klauspost, line 218:
```suggestion
	for nic, metrics := range other.ByNet {
```
Key is not a disk... Is it NIC? From `InterfaceName`?

metrics.go
klauspost: OK, so you are adding these by Host.

You should add data to `Metrics` struct and not here.

This will automatically populate `ByHost` since you are setting `ByHost: true`

metrics.go
```diff
@@ -206,6 +210,14 @@ func (r *RealtimeMetrics) Merge(other *RealtimeMetrics) {
 	for disk, metrics := range other.ByDisk {
 		r.ByDisk[disk] = metrics
 	}
+
+	// merge NetStats
```
klauspost, line 214: This should be in `func (r *Metrics) Merge(other *Metrics)`

harshavardhana (approved):

metrics.go
```diff
@@ -549,3 +558,24 @@ func (o *SiteResyncMetrics) Merge(other *SiteResyncMetrics) {
...
+	CollectedAt time.Time `json:"collected"`
+
+	// net of Interface
+	InterfaceName string `json:"interfaceName"`
+
+	NetStats procfs.NetDevLine `json:"netstats"`
+}
+
+// Merge other into 'o'.
+func (n *NetMetrics) Merge(other *NetMetrics) {
```
klauspost, line 573: This must *accumulate*, so new value of `n` represents `n + other`. This way we accumulate for all hosts in the total.

klauspost (changes requested):

klauspost (approved):
LGTM

PR #212, opened 2023-07-25
feat: support perf client

jiuker:
To support `mc support perf client`


perf-client.go
```diff
@@ -0,0 +1,95 @@
...
+	return n, nil
+}
+
+var _ io.Reader = &clientPerfReader{}
+
+// ClientPerf - perform net from client to MinIO servers
+func (adm *AdminClient) ClientPerf(ctx context.Context) (result ClientPerfResult, err error) {
+	defer func() {
+		result.Endpoint = adm.endpointURL.String()
+	}()
```
harshavardhana, line 75: why this defer? just set it to `result.Endpoint = adm.endpointURL.String()`

perf-client.go
```diff
@@ -0,0 +1,95 @@
...
+	queryVals := make(url.Values)
+	reader := &clientPerfReader{}
+	reader.Start()
+	_, err = adm.executeMethod(ctx, http.MethodPost, requestData{
+		queryValues:   queryVals,
+		relPath:       adminAPIPrefix + "/speedtest/client/devnull",
+		contentReader: reader,
+	})
+	reader.End()
```
harshavardhana, line 84: you are not passing any context, when does it end?

harshavardhana (changes requested):

perf-client.go
```diff
@@ -0,0 +1,133 @@
...
+	err = dec.Decode(&extraTime)
+	if err != nil {
+		return ClientPerfResult{}, err
+	}
+	durSpend := reader.endTime.Sub(reader.startTime).Nanoseconds()
+	if extraTime.Dur > 0 {
+		durSpend = durSpend - extraTime.Dur
+	}
+	if durSpend <= 0 {
+		return ClientPerfResult{}, fmt.Errorf("DurSpend is unexpected! Spent too much at get the lock!")
```
harshavardhana, line 127:
```suggestion
		return ClientPerfResult{}, fmt.Errorf("unexpected spent time duration, mostly NTP errors on the server")
```

perf-client.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return ClientPerfResult{
+		TX:    uint64(float64(reader.count) / time.Duration(durSpend).Seconds()),
```
harshavardhana, line 130:
```suggestion
		TX:    uint64(float64(reader.count) / time.Duration(durSpend).Seconds()),
```

Let's actually provide lower values from an API point of view such that `mc` can do humanize or pretty calculations.

```
ClientPerfResult{
   BytesRead: uint64(reader.count),
   TimeSpent: durSpend,
}
```

harshavardhana (changes requested):

harshavardhana (approved):

PR #213, opened 2023-08-03
add 'expire' batch job

harshavardhana:


trace.go
```diff
@@ -49,6 +49,10 @@ const (
 	TraceBatchReplication
 	// TraceBatchKeyRotation will trace batch keyrotation operations.
 	TraceBatchKeyRotation
+	// TraceBatchExpire will trace batch expiration operations.
+	TraceBatchExpire
+	// TraceBatch will trace all batch operations.
+	TraceBatch
```
krisis, line 55: We should define `TraceBatch` outside of this `iota` based expressions. 

`TraceBatch` is simply,
```go
const(
  TraceBatch = TraceBatchReplication|TraceBatchKeyRotation|TraceBatchExpire // |TraceBatch<NextFeature>
)
```

trace.go
```diff
@@ -49,6 +49,10 @@ const (
 	TraceBatchReplication
 	// TraceBatchKeyRotation will trace batch keyrotation operations.
 	TraceBatchKeyRotation
+	// TraceBatchExpire will trace batch expiration operations.
```
krisis, line 52: It is better to have newer flags appear after all the current flags. This way the enum value for these constants remain the same across madmin-go versions.

harshavardhana: the reason is it follow a more easier-to-read convention like `TraceBatch` so that we can avoid disjointed constants.

Also enum is compiled in so there is no problem as such.

krisis (approved):

PR #218, opened 2023-08-18
Add MetricsClient and NodeMetrics API

cesnietor:
This adds a new client `MetricsClient` to fetch Metrics Data from MinIO for the Prometheus endpoints.
It also includes the method `NodeMetrics` which is the equivalent to `v2/metrics/node` on MinIO.

The idea is to be able to use this in mc and other MinIO products for convenience and to avoid duplicated code. 
It was not included as part of the AdminClient since this is only for the metrics endpoints which requires a different and much simpler request flow. 
Based on how mc is currently calling it [here](https://github.com/minio/mc/blob/master/cmd/admin-prometheus-metrics.go#L74).
And on the current implementation of AdminClient [here](https://github.com/minio/madmin-go/blob/add-metrics-client/api.go#L50).

### Changes
- Added MetricsClient
- Added NodeMetrics method
- Added Unittests

### Test Steps:
- Create a client
- Run Metrics API
- Parse the response to string to visualize it

```
ctx := context.Background()
madminClient, _ := madmin.NewMetricsClient(
		"http://localhost:9000",
		"accesskey",
		"secretKey",
		false,
	)
nodeMetrics, _ := metricsClient.NodeMetrics(ctx)

result, _ := json.MarshalIndent(nodeMetrics, "", " ")
log.Println(string(result))
```
Output would look like:
```
[
 {
  "name": "go_memstats_alloc_bytes",
  "help": "Number of bytes allocated and still in use.",
  "type": "GAUGE",
  "metrics": [
   {
    "value": "1.76933784e+08"
   }
  ]
 },
 {
  "name": "go_memstats_frees_total",
  "help": "Total number of frees.",
  "type": "COUNTER",
  "metrics": [
   {
    "value": "2.59015547e+08"
   }
  ]
 },
 ...
 ]
```


go.mod
```diff
@@ -16,22 +16,33 @@ require (
...
 	github.com/tklauser/go-sysconf v0.3.11 // indirect
 	github.com/tklauser/numcpus v0.6.0 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
 	golang.org/x/sync v0.1.0 // indirect
 	golang.org/x/sys v0.10.0 // indirect
+	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
 )
 
+require (
```
cesnietor, line 39: Go gave me this warning and I ran `go mod tidy -go=1.16 && go mod tidy -go=1.17`:
```
// github.com/minio/madmin-go/v3 imports
// 	github.com/shirou/gopsutil/v3/cpu tested by
// 	github.com/shirou/gopsutil/v3/cpu.test imports
// 	github.com/stretchr/testify/assert loaded from github.com/stretchr/testify@v1.8.1,
// 	but go 1.16 would select v1.8.2
// github.com/minio/madmin-go/v3 imports
// 	github.com/minio/minio-go/v7/pkg/credentials imports
// 	github.com/json-iterator/go tested by
// 	github.com/json-iterator/go.test imports
// 	github.com/stretchr/testify/require loaded from github.com/stretchr/testify@v1.8.1,
// 	but go 1.16 would select v1.8.2

// To upgrade to the versions selected by go 1.16:
// 	go mod tidy -go=1.16 && go mod tidy -go=1.17
// If reproducibility with go 1.16 is not needed:
// 	go mod tidy -compat=1.17
// For other options, see:
// 	https://golang.org/doc/modules/pruning
```
Please let me know if we should do it differently

harshavardhana: You need to run go mod tidy -compat=1.20 - we do not support go1.16

go.mod
```diff
@@ -36,13 +41,4 @@ require (
...
-require (
-	github.com/golang-jwt/jwt/v4 v4.5.0
-	github.com/lufia/plan9stats v0.0.0-20230110061619-bbe2e5e100de // indirect
-	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	github.com/power-devops/perfstat v0.0.0-20221212215047-62379fc7944b // indirect
-	github.com/prometheus/client_model v0.4.0
-	github.com/prometheus/prom2json v1.3.3
-)
-
 replace github.com/minio/minio-go/v7 => github.com/poornas/minio-go/v7 v7.0.0-20230816201254-210e170bdc4c
```
harshavardhana, line 44: we need to remove this as well - is this added by you @poornas ?

harshavardhana (changes requested):

go.mod
cesnietor: yes it seems so, not sure if it was intentional and temporary.

harshavardhana: minio-go is released so remove this and use the latest release @cesnietor

cesnietor: done.

cesnietor:
@harshavardhana, @klauspost  told me it might not be the best idea to expose the NodeMetrics method in the client since once the client is behind a load balancer or something like that, it will fetch any node info. Which now it would not make much sense to add cluster and bucket but not node info. I'll check what we can do here.

harshavardhana:
> @harshavardhana, @klauspost told me it might not be the best idea to expose the NodeMetrics method in the client since once the client is behind a load balancer or something like that, it will fetch any node info. Which now, it makes me think, that it would not make much sense to add cluster and bucket but not node info. I'll check what we can do here.

Per node level scraping is part of Prometheus service monitoring @cesnietor you do not have to worry about that.

Each node specific endpoint must have its job defined in k8s and it is scraped independently. So similarly whoever needs access to node metrics must have access to all endpoints as well that can resolve and fetch from relevant node if needed.

This is a basic requirement for this API to work.

harshavardhana (approved):

reivaj05 (approved):

PR #220, opened 2023-08-23
feat:add fields for TTFB 

jiuker:
add field for TTFB


klauspost (commented):
What is up with all the removed dependencies?

perf-site-replication.go
```diff
@@ -63,3 +67,38 @@ func (adm *AdminClient) SiteReplicationPerf(ctx context.Context, duration time.D
...
+func (s *SiteNetPerfReader) Start() {
+	buf := make([]byte, 4*1024-1)
+	rand.Read(buf)
+	s.buf = buf
+	s.startTime = time.Now()
+}
+
+func (s *SiteNetPerfReader) Read(p []byte) (n int, err error) {
+	s.readCounter++
+	if s.readCounter == 1 {
```
harshavardhana, line 89: use switch case instead of double `if` @jiuker

perf-site-replication.go
```diff
@@ -63,3 +67,38 @@ func (adm *AdminClient) SiteReplicationPerf(ctx context.Context, duration time.D
...
+		n = copy(p, []byte{1})
+	} else if s.readCounter == 2 {
+		s.TimeToFirstByte = time.Now().Sub(s.startTime)
```
harshavardhana, line 92:
```suggestion
		s.TimeToFirstByte = time.Since(s.startTime)
```

perf-site-replication.go
```diff
@@ -63,3 +67,38 @@ func (adm *AdminClient) SiteReplicationPerf(ctx context.Context, duration time.D
...
+		n = copy(p, s.buf)
+	} else {
+		return 0, io.EOF
```
harshavardhana, line 95: You do not need to only send the first byte via this reader instead it must be the first Read() itself how much was received by the other end. 

Look at the server-side implementation of how a response recorder is used to calculate time to the first byte.

harshavardhana (changes requested):

harshavardhana (changes requested):
don't use only `go mod tidy` - please use `go mod tidy -compat=1.19`

jiuker:
@harshavardhana PTAL

PR #226, opened 2023-09-12
Add type for CPU metrics

anjalshireesh:
to include the CPU times and load stats

Also introduce following additional fields in MemInfo:
- Used
- Free
- Shared
- Cache
- Buffer

So that these can be used in upcoming resource metrics feature.


klauspost (approved):
LGTM

kannappanr (approved):

harshavardhana (approved):

PR #227, opened 2023-09-19
Add support of per bucket scanner report

vadmeste:
Each bucket will be scanned individually, this commit will deprecate 
old cycle scan metrics since there is no concept of a full cycle in the
upcoming per bucket scan change.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -234,14 +234,12 @@ type ScannerMetrics struct {
 	// Time these metrics were collected
 	CollectedAt time.Time `json:"collected"`
 
-	// Current scanner cycle
-	CurrentCycle uint64 `json:"current_cycle"`
+	CurrentCycle      uint64      `json:"current_cycle"`        // Deprecated Mar 2024
+	CurrentStarted    time.Time   `json:"current_started"`      // Deprecated Mar 2024
+	CyclesCompletedAt []time.Time `json:"cycle_complete_times"` // Deprecated Mar 2024
 
-	// Start time of current cycle
-	CurrentStarted time.Time `json:"current_started"`
-
-	// History of when last cycles completed
-	CyclesCompletedAt []time.Time `json:"cycle_complete_times"`
+	// Number of buckets currently scanning
+	OngoingBuckets int `json:"ongoing_buckets"`
 
 	// Number of accumulated operations by type since server restart.
 	LifeTimeOps map[string]uint64 `json:"life_time_ops,omitempty"`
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -223,19 +223,22 @@ func (r *RealtimeMetrics) Merge(other *RealtimeMetrics) {
...
 // ScannerMetrics contains scanner information.
 type ScannerMetrics struct {
 	// Time these metrics were collected
 	CollectedAt time.Time `json:"collected"`
 
-	// Current scanner cycle
-	CurrentCycle uint64 `json:"current_cycle"`
+	CurrentCycle      uint64      `json:"current_cycle"`        // Deprecated Sep 2023
+	CurrentStarted    time.Time   `json:"current_started"`      // Deprecated Sep 2023
+	CyclesCompletedAt []time.Time `json:"cycle_complete_times"` // Deprecated Sep 2023
```
klauspost, line 238: Why is this deprecated?

metrics.go
```diff
@@ -290,10 +293,16 @@ func (s *ScannerMetrics) Merge(other *ScannerMetrics) {
...
 		return
 	}
+
 	if s.CollectedAt.Before(other.CollectedAt) {
 		// Use latest timestamp
 		s.CollectedAt = other.CollectedAt
 	}
+
+	if s.BucketsScanInfo.OngoingBuckets < other.BucketsScanInfo.OngoingBuckets {
+		s.BucketsScanInfo.OngoingBuckets = other.BucketsScanInfo.OngoingBuckets
```
klauspost, line 303: So this is only being sent from the (master) server controlling the scan?

metrics.go
vadmeste: > Why is this deprecated?

I didn't see this; well, since a new bucket will be scanned when there is an available disk; there is no concept of scan cycle anymore; let's discuss this futher if you want

scanner.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"time"
+)
+
+type BucketScanInfo struct {
```
harshavardhana, line 29: Add some comments here

scanner.go
```diff
@@ -0,0 +1,62 @@
...
+	Pool, Set   int
+	Cycle       uint64
+	LastUpdate  time.Time
+	LastStarted time.Time
+	Completed   []time.Time
+}
+
+func (adm *AdminClient) BucketScanInfo(ctx context.Context, bucket string) ([]BucketScanInfo, error) {
```
harshavardhana, line 37: Same here

harshavardhana (approved):

PR #232, opened 2023-10-03
Add ResourceMetrics method

cesnietor:
Similar to https://github.com/minio/madmin-go/pull/218, this adds the Resource method equivalent to `/minio/v2/metrics/resource` endpoint.

Also renames a method to be more clear on what it does.

### Test Steps:
- Create a client
- Run Metrics API
- Parse the response to string to visualize it

```
ctx := context.Background()
madminClient, _ := madmin.NewMetricsClient(
		"http://localhost:9000",
		"accesskey",
		"secretKey",
		false,
	)
metrics, _ := metricsClient.ResourceMetrics(ctx)

result, _ := json.MarshalIndent(metrics, "", " ")
log.Println(string(result))
```
Output would look like:
```
[
 {
  "name": "someMetrics",
  "help": "Some help",
  "type": "GAUGE",
  "metrics": [
   {
    "value": "1.76933784e+08"
   }
  ]
 },
 {
  "name": "someMetrics",
  "help": "Some other help.",
  "type": "COUNTER",
  "metrics": [
   {
    "value": "2.59015547e+08"
   }
  ]
 },
 ...
 ]
```

### DIFF
--- a/prometheus_metrics.go
+++ b/prometheus_metrics.go
@@ -41,7 +41,7 @@ func (client *MetricsClient) NodeMetrics(ctx context.Context) ([]*prom2json.Fami
 	}
 
 	// Execute GET on /minio/v2/metrics/node
-	resp, err := client.executeRequest(ctx, reqData)
+	resp, err := client.executeGetRequest(ctx, reqData)
 	if err != nil {
 		return nil, err
 	}
@@ -61,7 +61,7 @@ func (client *MetricsClient) ClusterMetrics(ctx context.Context) ([]*prom2json.F
 	}
 
 	// Execute GET on /minio/v2/metrics/cluster
-	resp, err := client.executeRequest(ctx, reqData)
+	resp, err := client.executeGetRequest(ctx, reqData)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


prometheus_metrics.go
```diff
@@ -81,7 +81,27 @@ func (client *MetricsClient) BucketMetrics(ctx context.Context) ([]*prom2json.Fa
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	return parsePrometheusResults(io.LimitReader(resp.Body, metricsRespBodyLimit))
+}
+
+// ResourceMetrics - returns Resource Metrics in Prometheus format
+func (client *MetricsClient) ResourceMetrics(ctx context.Context) ([]*prom2json.Family, error) {
```
anjalshireesh, line 98: The code for each of these metrics functions seems almost same, except for the `relativePath`. It can be put into a common function, something like:

```go
// FetchMetrics - returns Metrics of given subsystem in Prometheus format
func (client *MetricsClient) FetchMetrics(ctx context.Context, subSys string) ([]*prom2json.Family, error) {
	reqData := metricsRequestData{
		relativePath: "/v2/metrics/" + subSys,
	}

	// Execute GET on /minio/v2/metrics/<subSys>
	resp, err := client.executeGetRequest(ctx, reqData)
	if err != nil {
		return nil, err
	}
	defer closeResponse(resp)

	if resp.StatusCode != http.StatusOK {
		return nil, httpRespToErrorResponse(resp)
	}

	return parsePrometheusResults(io.LimitReader(resp.Body, metricsRespBodyLimit))
}
```

anjalshireesh (approved):
LGTM. One suggestion to reduce code duplication.

prometheus_metrics.go
cesnietor: @anjalshireesh  I like it, I've done it, thanks.

PR #240, opened 2023-10-24
Capture network info in health report

anjalshireesh:
Introduce a new struct `NetInfo` to store the driver and firmware version of a network interface.

Also,

- Introduce a function to populate `NetInfo` for a given interface
- Add `Revision` field for a drive partition and capture it in the health diagnostics.

### DIFF
--- a/health.go
+++ b/health.go
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2015-2022 MinIO, Inc.
+// Copyright (c) 2015-2023 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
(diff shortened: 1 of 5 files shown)


health.go
```diff
@@ -242,48 +243,63 @@ type Partition struct {
...
 type Partitions struct {
 	NodeCommon
 
 	Partitions []Partition `json:"partitions,omitempty"`
 }
 
-func getDeviceModel(partDevice string) (string, error) {
-	var model string
-
+func getDeviceInfo(partDevice string) (model string, revision string, err error) {
```
shtripat, line 261: Do we want to create a struct for `DeviceInfo` itself so that latest we can add more attributes if needed?

klauspost (approved):
LGTM. What Shubhendu proposes sounds reasonable, but not critical.

health.go
anjalshireesh: @shtripat done, PTAL

shtripat (approved):

PR #241, opened 2023-10-26
Allow to set a custom Transport for MetricsClient

dvaldivia:
Allows the `MetricsClient` to set a custom `http.Transport` so that custom certificates can be supported.

Additionally makes all the methods for the metrics client to receive pointers as the Struct MetricsClient had methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation.

### DIFF
--- a/metrics_client.go
+++ b/metrics_client.go
@@ -99,7 +99,7 @@ func privateNewMetricsClient(endpointURL *url.URL, jwtToken string, secure bool)
 }
 
 // executeGetRequest - instantiates a Get method and performs the request
-func (client MetricsClient) executeGetRequest(ctx context.Context, reqData metricsRequestData) (res *http.Response, err error) {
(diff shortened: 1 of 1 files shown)


cesnietor (approved):
LGTM

harshavardhana:
Wait this PR is wrong @dvaldivia we do not want to encourage SetCustomTransport

harshavardhana:
convert the `metrics_client.go`  `NewMetricsClient(endpoint, opts)`  and take the Transport as part of `opts.Transport = transport`

dvaldivia:
I did it this way since the `madmin.Client` had something similar we could update all clients at the same time to that new pattern, however `madmin.Client`  opts doens't support transport yet @harshavardhana

harshavardhana:
> I did it this way since the `madmin.Client` had something similar we could update all clients at the same time to that new pattern, however `madmin.Client` opts doens't support transport yet @harshavardhana

Yeah new clients can use newer style @dvaldivia we have to move madmin.New()

PR #244, opened 2023-10-30
add support for AWS AssumeRole with tiering

harshavardhana:


tier-s3.go
```diff
@@ -23,14 +23,17 @@ package madmin
...
+	AccessKey          string `json:",omitempty"`
+	SecretKey          string `json:",omitempty"`
+	Bucket             string `json:",omitempty"`
+	Prefix             string `json:",omitempty"`
+	Region             string `json:",omitempty"`
+	StorageClass       string `json:",omitempty"`
+	AWSRole            bool   `json:",omitempty"`
+	AWSRoleARN         string `json:",omitempty"`
+	AWSRoleSessionName string `json:",omitempty"`
+	DurationSeconds    int    `json:",omitempty"`
```
poornas, line 36: this is not used?

harshavardhana: I think I need to add an option pattern for it.

harshavardhana:
PTAL @poornas @krisis

poornas (approved):

PR #252, opened 2023-12-04
Added structs for bucket throttle

shtripat:


harshavardhana (changes requested):
We do not need a new API simple extend SetBucketQuota() @shtripat

shtripat:
> We do not need a new API simple extend SetBucketQuota() @shtripat

IIUC, its like we add `concurrent-req-count` and `apis` options to `mc quota set` command, right?

quota-commands.go
```diff
@@ -40,13 +40,20 @@ func (t QuotaType) IsValid() bool {
...
 }
 
+// BucketThrottleRule holds a bucket throttle rule
+type BucketThrottleRule struct {
+	ConcurrentRequestsCount uint64   `json:"concurrentRequestsCount"` // indicates no of concurrent requests
+	APIs                    []string `json:"apis"`                    // indicates list of APIs
+}
+
 // BucketQuota holds bucket quota restrictions
 type BucketQuota struct {
```
harshavardhana, line 50: Bring a helper function that tells you if throttling is enabled or not

harshavardhana (changes requested):

PR #253, opened 2023-12-05
Add CPU count in metrics

anjalshireesh:
so that it can be used by the resource metrics collector to calculate "percentage" load (load * 100 / cpu-count)

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -640,6 +640,7 @@ type CPUMetrics struct {
 
 	TimesStat *cpu.TimesStat `json:"timesStat"`
 	LoadStat  *load.AvgStat  `json:"loadStat"`
+	Count     int            `json:"count"`
 }
 
 // Merge other into 'm'.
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

metrics.go
```diff
@@ -640,6 +640,7 @@ type CPUMetrics struct {
 
 	TimesStat *cpu.TimesStat `json:"timesStat"`
 	LoadStat  *load.AvgStat  `json:"loadStat"`
+	Count     int            `json:"count"`
```
harshavardhana, line 643:
```suggestion
	CPUCount     int            `json:"cpuCount"`
```

harshavardhana (changes requested):

metrics.go
anjalshireesh: Done

cesnietor:
> is that all we need? lol, is will this value be already normalized based on the number of cores and such?

ah I just saw the other pr in minio https://github.com/minio/minio/pull/18596

cesnietor (approved):

harshavardhana (approved):

PR #254, opened 2023-12-07
tier: Add support of service principal to Azure

vadmeste:
In order to support service principal account when configuring tiering to a Azure storage

### DIFF
--- a/tier-azure.go
+++ b/tier-azure.go
@@ -19,6 +19,8 @@
 
 package madmin
 
+import "errors"
+
 //go:generate msgp -file $GOFILE
 
 // ServicePrincipalAuth holds fields for a successful SP authentication with Azure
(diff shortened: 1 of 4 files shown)


tier-azure.go
```diff
@@ -30,11 +36,27 @@ type TierAzure struct {
 	Prefix       string `json:",omitempty"`
 	Region       string `json:",omitempty"`
 	StorageClass string `json:",omitempty"`
+
+	SPAuth ServicePrincipalAuth `json:",omitempty"`
+}
+
+func (ti TierAzure) IsServicePrincipalEnabled() bool {
```
harshavardhana, line 43:
```suggestion
func (ti TierAzure) IsSPAuthEnabled() bool {
```

harshavardhana (changes requested):

tier-azure.go
vadmeste: Done!

tier-azure.go
```diff
@@ -21,6 +21,12 @@ package madmin
 
 //go:generate msgp -file $GOFILE
 
+type ServicePrincipalAuth struct {
```
harshavardhana, line 24:
```suggestion
// ServicePrincipalAuth - add a comment
type ServicePrincipalAuth struct {
```

harshavardhana (changes requested):

krisis (approved):

PR #256, opened 2023-12-18
tier: Add validation to tier configuration

vadmeste:
Bonus, also add missing aws role web identity and role ARN when editing the tier

### DIFF
--- a/cgroup/nolinux.go
+++ b/cgroup/nolinux.go
@@ -24,6 +24,6 @@ package cgroup
 import "errors"
 
 // GetMemoryLimit - Not implemented in non-linux platforms
-func GetMemoryLimit(pid int) (limit uint64, err error) {
+func GetMemoryLimit(_ int) (limit uint64, err error) {
 	return limit, errors.New("Not implemented for non-linux platforms")
 }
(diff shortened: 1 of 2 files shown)


tier-azure.go
```diff
@@ -43,6 +43,26 @@ type TierAzure struct {
...
+
+	if t.SPAuth.TenantID == "" || t.SPAuth.ClientID == "" || t.SPAuth.ClientSecret == "" {
+		return errors.New("the tenant and the client ID and the secret are required")
+	}
+
+	return nil
+}
+
 // IsSPEnabled returns true if all SP related fields are provided
 func (ti TierAzure) IsSPEnabled() bool {
```
harshavardhana, line 67:
```suggestion
func (t TierAzure) IsSPEnabled() bool {
```

harshavardhana (changes requested):

harshavardhana (approved):

vadmeste:
Rewrote this as per @krisis ask; we are going to make the validation in one place, in the server side. Meanwhile, this PR will only contain the new fields in the tier config data structure.

krisis (approved):

PR #257, opened 2023-12-25
Make EnableSha256 in mc perf object test configurable

moting9:
current mc perf test  hardcode   DisableContentSha256 as true, this PR is to make it configurable by mc cmd. 
cmd/perf-tests.go
 97 @@ -95,7 +95,7 @@ func selfSpeedTest(ctx context.Context, opts speedTestOpts) (SpeedTestResult, er
 98         userMetadata[globalObjectPerfUserMetadata] = "true" // Bypass S3 API freeze
 99         popts := minio.PutObjectOptions{
100                 UserMetadata:         userMetadata,
101 -               **DisableContentSha256: true,**

### DIFF
--- a/perf-object.go
+++ b/perf-object.go
@@ -66,6 +66,7 @@ type SpeedtestOpts struct {
 	StorageClass string        // Choose type of storage-class to be used while performing I/O
 	Bucket       string        // Choose a custom bucket name while performing I/O
 	NoClear      bool          // Avoid cleanup after running an object speed test
+	EnableSha256 bool          // Enable calculating sha256 for uploads
 }
 
 // Speedtest - perform speedtest on the MinIO servers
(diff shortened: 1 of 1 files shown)


perf-object.go
```diff
@@ -59,13 +59,14 @@ type SpeedTestResult struct {
...
-	Bucket       string        // Choose a custom bucket name while performing I/O
-	NoClear      bool          // Avoid cleanup after running an object speed test
+	Size                 int           // Object size used in speed test
+	Concurrency          int           // Concurrency used in speed test
+	Duration             time.Duration // Total duration of the speed test
+	Autotune             bool          // Enable autotuning
+	StorageClass         string        // Choose type of storage-class to be used while performing I/O
+	Bucket               string        // Choose a custom bucket name while performing I/O
+	NoClear              bool          // Avoid cleanup after running an object speed test
+	DisableContentSha256 bool          // Disable calculating sha256 for uploads
```
harshavardhana, line 69: EnableSha256

harshavardhana: We need to keep the current behavior as is to avoid breaking change in behavior.

harshavardhana (changes requested):

harshavardhana (approved):

PR #258, opened 2024-01-02
Capture transparent hugepages config in health report

anjalshireesh:
Add a new node `thp-config` under `sys -> config` and inside it add values from following config files:

/sys/kernel/mm/transparent_hugepage/enabled
/sys/kernel/mm/transparent_hugepage/defrag
/sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none

ref: https://tip.golang.org/doc/gc-guide#Linux_transparent_huge_pages

### DIFF
--- a/health.go
+++ b/health.go
@@ -396,17 +396,6 @@ type XFSErrorConfig struct {
 	MaxRetries int    `json:"max_retries"`
 }
 
-// THPConfigs - stores transparent huge pages (THP) related configs
-type THPConfigs struct {
-	// /sys/kernel/mm/transparent_hugepage/enabled
-	Enabled string `json:"enabled"`
-	// /sys/kernel/mm/transparent_hugepage/defrag
-	Defrag string `json:"defrag"`
-	// /sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none
-	MaxPtesNone int    `json:"max_ptes_none"`
-	Error       string `json:"error,omitempty"`
-}
-
 // GetOSInfo returns linux only operating system's information.
 func GetOSInfo(ctx context.Context, addr string) OSInfo {
 	if runtime.GOOS != "linux" {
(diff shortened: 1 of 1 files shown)


vadmeste (commented):
@anjalshireesh maybe we can make this structure more generic; there are a lot of entries under '/sys/kernel/mm/transparent_hugepage/' and I think we may need all of them in the future. It shouldn't hurt to have a generic structure to contain many entries.

anjalshireesh:
> @anjalshireesh maybe we can make this structure more generic; there are a lot of entries under '/sys/kernel/mm/transparent_hugepage/' and I think we may need all of them in the future. It shouldn't hurt to have a generic structure to contain many entries.

@vadmeste have changed it to `map`, PTAL.

health.go
```diff
@@ -487,6 +489,33 @@ func readIntFromFile(filePath string) (num int, err error) {
 	return strconv.Atoi(strings.TrimSpace(string(data)))
 }
 
+func getTHPConfigs() map[string]interface{} {
```
klauspost, line 492: Any reason for not `map[string]string`?... At least then it can't be "anything".

health.go
```diff
@@ -487,6 +489,33 @@ func readIntFromFile(filePath string) (num int, err error) {
...
+	configs := map[string]interface{}{}
+	data, err := os.ReadFile("/sys/kernel/mm/transparent_hugepage/enabled")
+	if err != nil {
+		return map[string]interface{}{"error": err.Error()}
```
klauspost, line 496: Is there a reason to fail completely if one of them can't be read? Why no just add the error to the returned stuct...
```suggestion
		configs["enabled_error"] = err.Error()
```
... etc.

health.go
anjalshireesh: One of the configs being captured (`max_ptes_none`) is an integer, so thought of keeping it flexible. We can of course store the integer also in its string form and use `Atoi` when reading it in `health-analyzer` if required. Would that be a more preferred approach?

health.go
anjalshireesh: ok, will change to not fail on error.

anjalshireesh: this is done.

health.go
shtripat: Now we can return `map[string]string` from function.

anjalshireesh: ok, will change accordingly.

harshavardhana (approved):

health.go
```diff
@@ -487,6 +489,24 @@ func readIntFromFile(filePath string) (num int, err error) {
...
+func getTHPConfigs() map[string]string {
+	configs := map[string]string{}
+	captureTHPConfig(configs, "/sys/kernel/mm/transparent_hugepage/enabled", "enabled")
+	captureTHPConfig(configs, "/sys/kernel/mm/transparent_hugepage/defrag", "defrag")
+	captureTHPConfig(configs, "/sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none", "max_ptes_none")
+	return configs
+}
+
+func captureTHPConfig(configs map[string]string, filePath string, cfgName string) {
+	errFieldName := "cfgName" + "_error"
```
shtripat, line 501:
```suggestion
	errFieldName := cfgName + "_error"
```

shtripat (approved):
lgtm

PR #259, opened 2024-01-05
Make ParsePrometheusResults method public

cesnietor:
We'll be using this method in other minIO products so sharing this so that we don't have to duplicate code. It already has tests.

### DIFF
--- a/prometheus_metrics.go
+++ b/prometheus_metrics.go
@@ -71,10 +71,11 @@ func (client *MetricsClient) fetchMetrics(ctx context.Context, subSystem string)
 		return nil, httpRespToErrorResponse(resp)
 	}
 
-	return parsePrometheusResults(io.LimitReader(resp.Body, metricsRespBodyLimit))
+	return ParsePrometheusResults(io.LimitReader(resp.Body, metricsRespBodyLimit))
 }
 
-func parsePrometheusResults(reader io.Reader) (results []*prom2json.Family, err error) {
+// ParsePrometheusResults Returns Prometheus objects from string reader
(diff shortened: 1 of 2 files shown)


pjuarezd (approved):

prakashsvmx (approved):

harshavardhana (changes requested):
I am not sure why this is implemented this way, we should fix the implementation first.

```
go func() {
		defer close(errChan)
		err = prom2json.ParseReader(reader, mfChan)
		if err != nil {
			errChan <- err
		}
	}()
	for mf := range mfChan {
		results = append(results, prom2json.NewFamily(mf))
	}
	if err := <-errChan; err != nil {
		return nil, err
	}
	return results, nil
```

There is no reason to have a go-routine for a blocking call. 

Let's rewrite it like this clean and simple. 

```go
func ParsePrometheusResults(reader io.Reader) (results []*prom2json.Family, err error) {
        // We could do further content-type checks here, but the                                                                                                                                                   
        // fallback for now will anyway be the text format                                                                                                                                                         
        // version 0.0.4, so just go for it and see if it works.                                                                                                                                                   
        var parser expfmt.TextParser
        metricFamilies, err := parser.TextToMetricFamilies(reader)
        if err != nil {
                return nil, fmt.Errorf("reading text format failed: %v", err)
	}
        results = make([]*prom2json.Family, 0, len(metricFamilies))
	for _, mf := range metricFamilies {
		results = append(results, prom2json.NewFamily(mf))
	}
        return results, nil
}
```

harshavardhana:
@cesnietor ^^ PTAL

harshavardhana:
This is merged to the master with cleaner implementation.

PR #269, opened 2024-02-07
Fix logTrace type mismatch

klauspost:
Now matches https://github.com/minio/pkg/blob/main/logger/message/log/entry.go#L46

Fixes infinite loop in GetLogs.


dvaldivia (approved):
LGTM

harshavardhana (approved):

PR #281, opened 2024-05-23
Capture /proc/cmdline in health report

anjalshireesh:
So that any checks related to it (e.g. IOMMU configuration) can be performed subsequently.

### DIFF
--- a/health.go
+++ b/health.go
@@ -451,8 +451,9 @@ func GetSysConfig(_ context.Context, addr string) SysConfig {
 		limits, err := proc.Limits()
 		if err != nil {
 			sc.Error = "rlimit: " + err.Error()
+		} else {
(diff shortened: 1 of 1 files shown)


shtripat (approved):
lgtm

klauspost:
Would this capture env vars?  A bit concerned if command is `MINIO_ROOT_PASSWORD=admin123 minio server ....` would also make it through. If so, we should filter.

harshavardhana (approved):

harshavardhana:
> Would this capture env vars? A bit concerned if command is `MINIO_ROOT_PASSWORD=admin123 minio server ....` would also make it through. If so, we should filter.

its OS `/proc/cmdline` - what you are thinking is `/proc/$(pidof minio)/env` @klauspost

PR #282, opened 2024-05-23
Expand Service Account expiration validation

taran-p:
Allows service accounts to be created with an expiration that is zero or the unix epoch time, as both of these are treated as no expiry in the server

### DIFF
--- a/user-commands.go
+++ b/user-commands.go
@@ -375,14 +375,27 @@ func validateSADescription(desc string) error {
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

vadmeste (approved):
LGTM

(usually passing Expiration as nil to the server is enough to indicate we are not setting any expiration but I see that this can cause issues with older versions of MinIO server

PR #288, opened 2024-07-01
heal: Add pool/set indexes to select a specific ES to heal

vadmeste:
A nil pool means select all pools
A nil set means select all sets

### DIFF
--- a/heal-commands.go
+++ b/heal-commands.go
@@ -54,8 +54,10 @@ type HealOpts struct {
 	UpdateParity bool         `json:"updateParity"` // Update the parity of the existing object with a new one
 	NoLock       bool         `json:"nolock"`
 
+	// Pool to heal. nil indicates "all pools" (and sets).
 	Pool *int `json:"pool,omitempty"`
-	Set  *int `json:"set,omitempty"`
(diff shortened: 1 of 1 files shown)


heal-commands.go
```diff
@@ -53,6 +53,9 @@ type HealOpts struct {
 	ScanMode     HealScanMode `json:"scanMode"`
 	UpdateParity bool         `json:"updateParity"` // Update the parity of the existing object with a new one
 	NoLock       bool         `json:"nolock"`
+
+	Pool *int `json:"pool,omitempty"`
+	Set  *int `json:"set,omitempty"`
```
klauspost, line 58:
```suggestion
	// Pool to heal. nil indicates "all pools" (and sets).
	Pool *int `json:"pool,omitempty"`
	// Set to heal. nil indicates "all sets". Should always be nil if Pool is nil.
	Set  *int `json:"set,omitempty"`
```

klauspost (changes requested):
Is this what you intend?

vadmeste:
> Is this what you intend?

correct

harshavardhana (approved):

klauspost (approved):

PR #292, opened 2024-07-25
Just make it MaxPingDur

klauspost:
I don't want to add expensive windowing. Keep it simple.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -690,7 +690,7 @@ type RPCMetrics struct {
 	OutQueue         int       `json:"outQueue"`
 	LastPongTime     time.Time `json:"lastPongTime"`
 	LastPingMS       float64   `json:"lastPingMS"`
-	Max1MinPingMS    float64   `json:"max1MinPingMS"`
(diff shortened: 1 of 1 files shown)


metrics.go
```diff
@@ -690,7 +690,7 @@ type RPCMetrics struct {
 	OutQueue         int       `json:"outQueue"`
 	LastPongTime     time.Time `json:"lastPongTime"`
 	LastPingMS       float64   `json:"lastPingMS"`
-	Max1MinPingMS    float64   `json:"max1MinPingMS"`
+	MaxPingDurMS     float64   `json:"maxPingDurMS"` // Maximum across all merged entries.
```
harshavardhana, line 693:
```suggestion
	MaxPingDurMS     float64   `json:"maxPingDurMS"` // Maximum across all merged RPC messages.
```

klauspost: No. It is across metric entries. Ping is not related to RPC messages - it is separate ping/pong roundtrip on the connection.

harshavardhana (approved):

PR #293, opened 2024-07-26
Use typed backend information

ramondeklein:
The `MinioInfo` structure contains backend information, but it's not typed.

PS: I'm not sure if there is a specific reason why this wasn't typed, so please close and comment if it cannot be typed for some reason.


harshavardhana:
> PS: I'm not sure if there is a specific reason why this wasn't typed, so please close and comment if it cannot be typed for some reason.

yeah its historic problem we had to change the type and ended up using interface{} because that older struct by mistake was not versioned properly.

I think we are keeping it here for really old deployments.

ramondeklein:
Okay, then I’ll close it.

PR #294, opened 2024-07-29
postgresql events notification

romanchechyotkin:
added functionality for creating postgresql events notification

### DIFF
--- a/examples/set_postgresql_notification.go
+++ b/examples/set_postgresql_notification.go
@@ -1,0 +1,58 @@
+//go:build ignore
+// +build ignore
+
+// Copyright (c) 2015-2022 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


romanchechyotkin:
image

romanchechyotkin:
then i want to add functionality for connecting configuared event notification to bucket in [minio-go](https://github.com/minio/minio-go) 
image

romanchechyotkin:
i suppose there is need to write kind of this functions for oher integrations like MySQl, Redis, Kafka and etc

harshavardhana (changes requested):
what purpose does this API solve that you can't do via `SetConfigKV` ?

romanchechyotkin:
using options struct instead of the raw string

klauspost:
Pro: Convenient
Cons: Seems a bit strange to offer this as a single config to abstract.

Undecided.

romanchechyotkin:
Yea, i am going to implement this functionality for all supported integratioins, theres is reused config values 
```go
        // Staging directory for undelivered messages e.g. '/home/events'
	QueueDir string

	// Maximum limit for undelivered messages, defaults to '10000'
	QueueLimit uint

	// Specify a comment to associate with the Postgresql configuration
	Comment string
```
The question is if it would be accepted as test for postgresql, if yes i continue to implement this functionality

harshavardhana:
It is not implemented purposely, so we don't have to bind by some structured contract. If we merge and publish this, we must maintain it for eternity. the KV API is kept fluid as the MinIO server changes it also changes automatically without making any new "madmin-go" releases. 

If you want, you can keep a fork for your work that wraps on top of this SDK.  While convenient there, this PR is a hidden burden on why we didn't do it ourselves.

PR #297, opened 2024-08-12
heal: Add Finished bool field to drive healing info

vadmeste:
Currently .healing.bin is removed when healing is finished. However, we are 
going to preserve .healing.bin to keep historical information. So adding 
Finished as a boolean flag to indicate that healing of a drive is finished 
will be needed.


klauspost (approved):
I assume that there is no way the current functionality would return it after it finished. So it seem that it would always make sense.

harshavardhana (approved):

PR #301, opened 2024-08-30
tier: Add force flag to remove a tier

vadmeste:
Sometimes, the tier is down or its address change and we could have 
many reasons why it is not possible to reach the tier server during removal 
of a tier. For that, we need to add a force flag.

### DIFF
--- a/tier.go
+++ b/tier.go
@@ -177,6 +177,40 @@ func (adm *AdminClient) RemoveTier(ctx context.Context, tierName string) error {
 	return nil
 }
 
+// RemoveTierOpts - options for a remote tiering removal
+type RemoveTierOpts struct {
+	Force bool
+}
+
+// RemoveTierV2 removes an empty tier identified by tierName, the tier is not
+// required to be reachable or empty if force flag is set to true.
+func (adm *AdminClient) RemoveTierV2(ctx context.Context, tierName string, opts RemoveTierOpts) error {
+	if tierName == "" {
+		return ErrTierNameEmpty
+	}
+
+	queryVals := url.Values{}
+	queryVals.Set("force", strconv.FormatBool(opts.Force))
+
+	reqData := requestData{
+		relPath:     path.Join(adminAPIPrefix, tierAPI, tierName),
+		queryValues: queryVals,
+	}
+
+	// Execute DELETE on /minio/admin/v3/tier/tierName to remove an empty tier.
+	resp, err := adm.executeMethod(ctx, http.MethodDelete, reqData)
+	defer closeResponse(resp)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusNoContent {
+		return httpRespToErrorResponse(resp)
+	}
+
+	return nil
+}
+
 // VerifyTier verifies tierName's remote tier config
 func (adm *AdminClient) VerifyTier(ctx context.Context, tierName string) error {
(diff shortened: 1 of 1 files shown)


harshavardhana:
what are the consequences of doing this wrong?

vadmeste:
> what are the consequences of doing this wrong?

for now, it is not possible to re-add the tier again. Currently if the tier contains some data, then the removal will not work until the user removes the data manually from the tier. However, there are some other corner cases, like the tier is down (like a user decided to kill the tiering itself). You can never remove this tier configuration in MinIO side in that case.

harshavardhana:
> for now, it is not possible to re-add the tier again. Currently if the tier contains some data, then the removal will not work until the user removes the data manually from the tier. However, there are some other corner cases, like the tier is down (like a user decided to kill the tiering itself). You can never remove this tier configuration in MinIO side in that case.

yes but this is disastrous, all the data that exists on the hot-tier will become stale and dangling. 

I don't think we need this feature at all.

vadmeste:
> I don't think we need this feature at all.

Well, it is a force flag, we can require --dangerous in addition to --force in mc. 

`mc ilm tier remove <ALIAS> AZTIER --force --dangerous`

I said there are some scenarios where you cannot remove the existing tier anymore, like a user who shutdowns the tier before removing the tier config in MinIO, the tier configuration will be stuck in MinIO forever.

harshavardhana:
> > I don't think we need this feature at all.
> 
> Well, it is a force flag, we can require --dangerous in addition to --force in mc.
> 
> `mc ilm tier remove <ALIAS> AZTIER --force --dangerous`
> 
> I said there are some scenarios where you cannot remove the existing tier anymore, like a user who shutdowns the tier before removing the tier config in MinIO, the tier configuration will be stuck in MinIO forever.

Then we can make this a hidden feature in mc, there is no good reason user should do that.

It's a data loss scenario and it's not a daily use behavior.

tier.go
```diff
@@ -155,12 +155,17 @@ func (adm *AdminClient) EditTier(ctx context.Context, tierName string, creds Tie
 }
 
 // RemoveTier removes an empty tier identified by tierName
-func (adm *AdminClient) RemoveTier(ctx context.Context, tierName string) error {
+func (adm *AdminClient) RemoveTier(ctx context.Context, tierName string, force bool) error {
```
harshavardhana, line 158: Please don't break existing APIs write new ones.

harshavardhana (changes requested):

tier.go
```diff
@@ -177,6 +177,35 @@ func (adm *AdminClient) RemoveTier(ctx context.Context, tierName string) error {
 	return nil
 }
 
+// RemoveTierV2 removes an empty tier identified by tierName, the tier is not
+// required to be reachable or empty  if force flag is set to true.
+func (adm *AdminClient) RemoveTierV2(ctx context.Context, tierName string, force bool) error {
```
harshavardhana, line 182:
```suggestion
func (adm *AdminClient) RemoveTierV2(ctx context.Context, opts TierOptions) error {
```

This is better for future extensibility.

harshavardhana (changes requested):

tier.go
vadmeste: what about now ? (little different to what you suggested)

harshavardhana (approved):

PR #305, opened 2024-09-10
Adding a subsystem for the new event_queue mechanism

zveinn:
Adding the new global queue as a configurable subsystem

### DIFF
--- a/parse-config.go
+++ b/parse-config.go
@@ -30,32 +30,32 @@ import (
 
 // Top level configuration key constants.
 const (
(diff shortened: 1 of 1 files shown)


harshavardhana:
what does the new subsystem do?

zveinn:
> what does the new subsystem do?

This is the sybsystem that handles the global audit event queuing.

klauspost (approved):

klauspost:
@zveinn Not needed any more?

PR #308, opened 2024-09-20
Adding the audit_event_queue sub system

zveinn:
This subsystem is used to manage the global audit queue. 
See more here: https://github.com/miniohq/eos/pull/76


parse-config.go
```diff
@@ -111,6 +112,7 @@ var SubSystems = set.CreateStringSet(
 	NotifyWebhookSubSys,
 	LambdaWebhookSubSys,
 	BrowserSubSys,
+	AuditEventQueueSubSys,
```
harshavardhana, line 115: create a new variable

```
EOSSubsystems
```

and then add previous Subsystems and this new one, otherwise upstream pulling this would automatically show up this subsystem incorrectly.

harshavardhana (changes requested):

parse-config.go
zveinn: gotcha, will refactor

harshavardhana (approved):

PR #323, opened 2024-12-27
Use HTTP client for fetching credentials

ramondeklein:
Follow-up to https://github.com/minio/minio-go/pull/2041 to also enable `AdminClient` to use the specified HTTP client for fetching credentials.


ramondeklein:
@harshavardhana We may want to use an official release of https://github.com/minio/minio-go instead of referencing a specific commit...

klauspost (approved):

harshavardhana:
> @harshavardhana We may want to use an official release of https://github.com/minio/minio-go instead of referencing a specific commit...

its okay we will update it in time, these are our projects only :-)

harshavardhana (approved):

PR #325, opened 2025-01-03
Add 'edition' field to cluster info related structs

anjalshireesh:
This will help identifying aistor release in
- mc admin info
- mc license register
- mc support diag

### DIFF
--- a/health.go
+++ b/health.go
@@ -996,6 +996,7 @@ type ServerInfo struct {
 	RuntimeVersion string            `json:"runtime_version"`
 	GCStats        *GCStats          `json:"gc_stats,omitempty"`
 	MinioEnvVars   map[string]string `json:"minio_env_vars,omitempty"`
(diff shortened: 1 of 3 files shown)


health.go
```diff
@@ -996,6 +996,7 @@ type ServerInfo struct {
 	RuntimeVersion string            `json:"runtime_version"`
 	GCStats        *GCStats          `json:"gc_stats,omitempty"`
 	MinioEnvVars   map[string]string `json:"minio_env_vars,omitempty"`
+	IsAIStor       *bool             `json:"is_aistor,omitempty"`
```
klauspost, line 999:
```suggestion
	Edition       string             `json:"edition"`
```

klauspost (commented):
How about just making it a text field?

health.go
anjalshireesh: done

harshavardhana (approved):

PR #333, opened 2025-02-18
Add v4 Admin APIs, older APIs move to v3

harshavardhana:
all deprecated APIs and structs are removed.

This SDK is now officially moving to v4
in-line with changes on the server side.


profiling-commands.go
```diff
@@ -47,77 +45,6 @@ const (
...
-type StartProfilingResult struct {
-	NodeName string `json:"nodeName"`
-	Success  bool   `json:"success"`
-	Error    string `json:"error"`
-}
-
-// StartProfiling makes an admin call to remotely start profiling on a
-// standalone server or the whole cluster in case of a distributed setup.
-//
-// Deprecated: use Profile API instead
```
klauspost, line 61: This will be extremely annoying for warp, since it will be unable to terminate benchmarks when using "autoterm" or otherwise interrrupted. It really need separate start/stop commands.

`--serverprof=cpu,mem` is extremely convenient.

harshavardhana: okay then it is not a deprecated API, will keep it back.

taran-p (approved):

PR #335, opened 2025-02-21
Add SRSessionPolicy type that unmarshalls correctly

taran-p:
helps fix miniohq/eos#444

### DIFF
--- a/cluster-commands_test.go
+++ b/cluster-commands_test.go
@@ -1,0 +1,52 @@
+package madmin
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestSRSessionPolicy_MarshalUnmarshalJSON(t *testing.T) {
+	tests := []struct {
+		name        string
+		policyBytes []byte
+	}{
+		{
+			name:        "ValidPolicy",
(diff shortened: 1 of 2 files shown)


taran-p:
@harshavardhana Is this what you were looking for? I'm merging into v3 because that's what I tested as, but could change to v4 if that would be better

harshavardhana:
> @harshavardhana Is this what you were looking for? I'm merging into v3 because that's what I tested as, but could change to v4 if that would be better

correct, you can add a unit test as well. Yeah v3 branch is fine for now, since v4 is going through some changes can't make a release yet.

cluster-commands_test.go
```diff
@@ -0,0 +1,52 @@
+package madmin
```
harshavardhana, line 1:
```suggestion
package madmin
```

preserve the copyright @taran-p

harshavardhana (changes requested):

cluster-commands_test.go
taran-p: whoops, fixed

cluster-commands_test.go
```diff
@@ -1,3 +1,22 @@
+//
+// Copyright (c) 2015-2024 MinIO, Inc.
```
harshavardhana:
```suggestion
// Copyright (c) 2015-2025 MinIO, Inc.
```

harshavardhana (approved):

PR #338, opened 2025-02-28
Add support for catalog batch job type

donatello:
- Add yaml template
- Add ListBatchJobTypes API
- Also add catalog metrics type

The ListBatchJobTypes API is basically a head request to the existing list jobs API. The existing list jobs APi in the server is updated to include a header  listing the types of jobs supported as a header (only for EOS).

### DIFF
--- a/batch-job.go
+++ b/batch-job.go
@@ -37,6 +37,7 @@ const (
 	BatchJobReplicate BatchJobType = "replicate"
 	BatchJobKeyRotate BatchJobType = "keyrotate"
 	BatchJobExpire    BatchJobType = "expire"
+	BatchJobCatalog   BatchJobType = "catalog"
 )
 
 // SupportedJobTypes supported job types
@@ -44,7 +45,7 @@ var SupportedJobTypes = []BatchJobType{
 	BatchJobReplicate,
 	BatchJobKeyRotate,
 	BatchJobExpire,
-	// add new job types
+	// No need to add new types here, they are added directly in server.
 }
 
 // BatchJobReplicateTemplate provides a sample template
@@ -431,4 +432,26 @@ func (adm *AdminClient) CancelBatchJob(ctx context.Context, jobID string) error 
 		return httpRespToErrorResponse(resp)
 	}
 	return nil
+}
+
+// CatalogDataFile contains information about an output file from a catalog job run.
+type CatalogDataFile struct {
+	Key         string `json:"key"`
+	Size        uint64 `json:"size"`
+	MD5Checksum string `json:"MD5Checksum"`
+}
+
+// CatalogManifestVersion represents the version of a catalog manifest.
+type CatalogManifestVersion string
+
+const (
+	CatalogManifestVersionV1 CatalogManifestVersion = "v1"
+)
+
+// CatalogManifest represents the manifest of a catalog job's result.
+type CatalogManifest struct {
+	Version        CatalogManifestVersion `json:"version"`
+	JobID          string                 `json:"jobID"`
+	StartTimestamp string                 `json:"startTimestamp"`
+	Files          []CatalogDataFile      `json:"files"`
 }
--- a/metrics.go
+++ b/metrics.go
@@ -489,6 +489,7 @@ type JobMetric struct {
 	Replicate *ReplicateInfo   `json:"replicate,omitempty"`
 	KeyRotate *KeyRotationInfo `json:"rotation,omitempty"`
 	Expired   *ExpirationInfo  `json:"expired,omitempty"`
+	Catalog   *CatalogInfo     `json:"catalog,omitempty"`
 }
 
 type ReplicateInfo struct {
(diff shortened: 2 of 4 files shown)


batch-job.go
```diff
@@ -349,6 +415,38 @@ func (adm *AdminClient) ListBatchJobs(ctx context.Context, fl *ListBatchJobsFilt
 	return result, nil
 }
 
+// SupportedBatchJobsHeader is the header key for supported batch jobs included
+// by the server in the list batch jobs response.
+const SupportedBatchJobsHeader = "X-MinIO-Supported-Batch-Jobs"
+
+// ListBatchJobTypes lists the supported batch job types.
+func (adm *AdminClient) ListBatchJobTypes(ctx context.Context) ([]string, error) {
```
donatello, line 423: This is used to check if server supported catalog job type

batch-job.go
```diff
@@ -37,16 +38,36 @@ const (
 	BatchJobReplicate BatchJobType = "replicate"
 	BatchJobKeyRotate BatchJobType = "keyrotate"
 	BatchJobExpire    BatchJobType = "expire"
+	BatchJobCatalog   BatchJobType = "catalog"
 )
 
 // SupportedJobTypes supported job types
+//
+// Deprecated: Use ClientSupportedJobTypes instead.
```
klauspost, line 46: What is the point of deprecating this? 

I guess it also comes down to that I don't understand what "client" refers to in `ClientSupportedJobTypes`.

batch-job.go
```diff
@@ -349,6 +423,38 @@ func (adm *AdminClient) ListBatchJobs(ctx context.Context, fl *ListBatchJobsFilt
 	return result, nil
 }
 
+// SupportedBatchJobsHeader is the header key for supported batch jobs included
+// by the server in the list batch jobs response.
+const SupportedBatchJobsHeader = "X-MinIO-Supported-Batch-Jobs"
```
klauspost, line 428:
```suggestion
const SupportedBatchJobsHeader = "X-Minio-Supported-Batch-Jobs"
```
(canonical)

batch-job.go
```diff
@@ -349,6 +423,38 @@ func (adm *AdminClient) ListBatchJobs(ctx context.Context, fl *ListBatchJobsFilt
...
+			// This is the status code returned by community MinIO server (which
+			// does not support HEAD request for this API endpoint). In this
+			// case, we return an empty list of batch job types.
+			return []string{}, nil
+		}
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	supportedBatchJobs := resp.Header.Get(SupportedBatchJobsHeader)
+	supportedBatchJobTypes := strings.Split(supportedBatchJobs, ",")
```
klauspost, line 453: Should we validate them against the ones we know? Or is it fine to just return whatever is supported?

(optionally, you can make the a `func (b BatchJobType) Known() bool` that returns if the sdk knows the type)

batch-job.go
```diff
@@ -191,6 +212,57 @@ const BatchJobExpireTemplate = `expire:
...
+
+  # "versions" specifies if only current or all versions of each object should
+  # be processed.
+  versions: current # Other valid value is "all"
+
+  # The list of optional output fields to include. Please refer to documentation
+  # for the full list of default and optional fields available.
+  includeFields:
+    - ETag
+    - IsMultipartUpload
```
klauspost, line 260: Checksum? Tags? Those would be really nice to have (feel free to close, since this is pretty irrelevant here)

batch-job.go
```diff
@@ -349,6 +423,38 @@ func (adm *AdminClient) ListBatchJobs(ctx context.Context, fl *ListBatchJobsFilt
 	return result, nil
 }
 
+// SupportedBatchJobsHeader is the header key for supported batch jobs included
+// by the server in the list batch jobs response.
+const SupportedBatchJobsHeader = "X-MinIO-Supported-Batch-Jobs"
+
+// ListBatchJobTypes lists the supported batch job types.
+func (adm *AdminClient) ListBatchJobTypes(ctx context.Context) ([]string, error) {
```
klauspost, line 431: This is what we are returning, right?

```suggestion
func (adm *AdminClient) ListBatchJobTypes(ctx context.Context) ([]BatchJobType, error) {
```

metrics.go
```diff
@@ -523,6 +524,22 @@ type KeyRotationInfo struct {
...
 }
 
+type CatalogInfo struct {
+	LastBucketScanned string `json:"lastBucketScanned"`
+	LastObjectScanned string `json:"lastObjectScanned"`
+	LastBucketMatched string `json:"lastBucketMatched"`
+	LastObjectMatched string `json:"lastObjectMatched"`
+
+	ObjectsScannedCount uint64 `json:"objectsScannedCount"`
+	ObjectsMatchedCount uint64 `json:"objectsMatchedCount"`
```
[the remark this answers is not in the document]
donatello: A catalog job scans only a single bucket in a particular job instance.

batch-job.go
```diff
@@ -191,6 +204,57 @@ const BatchJobExpireTemplate = `expire:
...
+  #
+  # once: starts immediately and runs only once
+  #
+  # All the following schedules do not start the job immediately on submission.
+  # daily: runs every day at roughly the same time
+  # weekly: runs every sunday at roughly the same time
+  # monthly: runs on first sunday of every month.
+  # yearly: runs on first sunday of January every year.
+	#
+  # FIXME: only once supported for initial implementation
```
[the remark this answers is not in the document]
donatello: Fixed.

batch-job.go
donatello: Just that it is ambiguously named. With this change, catalog job is only going to be supported by EOS and not open-source minio. So `ClientSupportedJobTypes` is going to be used only in mc and other client applications.

batch-job.go
donatello: Yeah, the full list will be added in docs. The YAML file template already looks quite large and more will be added.

batch-job.go
donatello: Yes that is checked against `ClientSupportedJobTypes` by client app code (so they could easily display an error instead of the unsupported ones being filtered out by this library).

batch-job.go
```diff
@@ -191,6 +204,57 @@ const BatchJobExpireTemplate = `expire:
...
+  #     acceptable. For example: storage usage
+  #   "strict" -> use more resources and complete slower, but avoids returning
+  #     stale data.
+  # In either case, objects modified (created, replaced, removed) during the
+  # catalog job run may not be included in the output.
+  mode: fast
+
+  # "versions" specifies if only current or all versions of each object should
+  # be processed.
+  versions: current # Other valid value is "all"
```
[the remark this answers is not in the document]
donatello: I that's we intend with `versions` already.

harshavardhana:
My opinion lets move this to v4 @donatello i.e on the main branch and avoid adding "deprecation notices" and bring new things to v3

klauspost (approved):
Disagree on the need to change the field name - it doesn't really disambiguate - and just seems like a "feel-good" change. But I can live with it.

donatello:
Will be resending this as it has changed significantly now

PR #339, opened 2025-03-04
Creating extended erasure set info 

zveinn:
Extended erasure set info will be used by minio to give use basic information on the Erasure set plus some extended information. Right now it's only a list of disks but we might add something more later as we develop these apis.


klauspost (commented):
LGTM. But I feel it will be somewhat contentious that we return binary data. Maybe ask around - it will surely be more efficient than JSON.

harshavardhana (approved):

PR #342, opened 2025-03-08
Add `RevokeTokens` and `RevokeTokensLDAP` APIs

taran-p:
Add `RevokeTokens` and `RevokeTokensLDAP` APIs, as well as user provider type constants.

miniohq/eos#498


harshavardhana (approved):

user-commands.go
```diff
@@ -760,3 +760,70 @@ func (adm *AdminClient) TemporaryAccountInfo(ctx context.Context, accessKey stri
 	}
 	return infoResp, nil
 }
+
+// User provider types
+const (
+	InternalProvider    = "internal"
```
donatello, line 766: Would it be better to call this as `builtin` instead of `internal` (the latter sounds like it refers to something to internal to minio system's functioning/implementation)?

taran-p: True, that is what we call them for API calls as well

donatello (approved):

PR #343, opened 2025-03-11
Add license to server info

anjalshireesh:
So that it can be captured in health diagnostics and customer can be alerted when license is about to expire.


shtripat (approved):
lgtm

harshavardhana (approved):

PR #344, opened 2025-03-11
Add GenerateBatchJobV2 + GetSupportedBatchJobTypes

donatello:
This function calls the server's job template generation API endpoint.

### DIFF
--- a/batch-job.go
+++ b/batch-job.go
@@ -58,7 +58,7 @@ const BatchJobReplicateTemplate = `replicate:
     prefix: PREFIX # 'PREFIX' is optional
     # If your source is the 'local' alias specified to 'mc batch start', then the 'endpoint' and 'credentials' fields are optional and can be omitted
     # Either the 'source' or 'remote' *must* be the "local" deployment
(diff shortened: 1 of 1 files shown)


batch-job.go
```diff
@@ -305,6 +306,73 @@ func (adm *AdminClient) GenerateBatchJob(_ context.Context, opts GenerateBatchJo
 	return "", fmt.Errorf("unknown batch job requested: %s", opts.Type)
 }
 
+// SupportedBatchJobsHeader is the header key for supported batch jobs included
+// by the server in the list batch jobs response.
+const SupportedBatchJobsHeader = "X-Minio-Supported-Batch-Jobs"
```
harshavardhana, line 311: why have this kind of an API? why not get a JSON response?

donatello: It was to save an additional request (to list supported ones) when server reports that a job is not supported on generate, but it's probably cleaner to do a separate API. Will update.

donatello:
PTAL @harshavardhana @krisis

vadmeste (approved):

krisis (approved):

PR #348, opened 2025-03-20
Add OpenTelemetry types and handlers

klauspost:
Allows for streaming metrics from AIStor servers.


opentelemetry.go
```diff
@@ -0,0 +1,119 @@
+package madmin
```
harshavardhana, line 1: Add missing copyright headed..

harshavardhana (changes requested):

zveinn (approved):

harshavardhana (approved):

PR #349, opened 2025-03-24
adding object-info method

zveinn:


info-v4-object.go
```diff
@@ -0,0 +1,145 @@
...
+// It contains basic information on a part and it's xl.meta content.
+type ObjectInspectPart struct {
+	Part            int
+	Pool            int
+	Host            string
+	Set             int
+	Drive           string
+	Filename        string
+	XLVersion       xlMetaV2Version
+	XLVersionHeader xlMetaV2VersionHeader
```
klauspost, line 83: These types not being exported as a bit annoying since you can't declare them when using this package.

info-v4-object.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	bb, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	files = make(map[string]*ObjectInspectInfo)
+	err = json.Unmarshal(bb, &files)
```
klauspost, line 65:
```suggestion
	dec := json.NewDecoder(resp.Body)
	err = dec.Decode(&files)
```

info-v4-object.go
```diff
@@ -0,0 +1,145 @@
...
+	Name    string    `json:"name"`
+	Dir     bool      `json:"dir"`
+	Mode    uint32    `json:"mode"`
+}
+
+type xlMetaV2Version struct {
+	Type             int                   `json:"Type"`
+	ObjectV2         *xlMetaV2Object       `json:"V2Obj,omitempty"`
+	DeleteMarker     *xlMetaV2DeleteMarker `json:"DelObj,omitempty"`
+	WrittenByVersion uint64                `msg:"v"`
```
klauspost, line 117:
```suggestion
	WrittenByVersion uint64                `json:"v"`
```

klauspost (commented):
Just some stuff we might as well get now so it doesn't bite us later.

info-v4-object.go
zveinn: Why do we need a decoder ? it's not a streaming response

klauspost (approved):
<3

zveinn:
do not merge yet, i'm making some adjustments

zveinn:
moving this to v3

PR #350, opened 2025-03-25
Implementing the Object Summary API

zveinn:


klauspost (approved):

summary-v4-object.go
```diff
@@ -0,0 +1,166 @@
...
+// ObjectSummary calls minio to search for all files and parts
+// related to the given object, across all disks.
+func (adm *AdminClient) ObjectSummary(ctx context.Context, objOpts ObjectSummaryOptions) (objectSummaryResponse *ObjectSummaryRespose, err error) {
+	form := make(url.Values)
+	if objOpts.Bucket != "" {
+		form.Add("bucket", objOpts.Bucket)
+	}
+	if objOpts.Object != "" {
+		form.Add("object", objOpts.Object)
+	}
```
harshavardhana, line 45: what is the behavior when bucket is not provided? or object is not provided?

summary-v4-object.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, httpRespToErrorResponse(resp)
+	}
+
+	objectSummaryResponse = new(ObjectSummaryRespose)
+	dec := json.NewDecoder(resp.Body)
```
harshavardhana, line 63: why do we care about JSON? can we get this in msgpack instead?

harshavardhana: worried if this becomes too big, because you are also using 

```
	Files map[string]*ObjectSummaryInfo
```

which can be un-ending amount of files if someone tries to get this information across namespace.

harshavardhana (changes requested):

klauspost (approved):

summary-v4-object.go
zveinn: The idea was to enable multiple object support later, it's essentially the same kind of "danger" as inspect object. But thinking about it, users will be using this much so I should probably limit this to a single object for now and make sure it won't just walk infinite depth

summary-v4-object.go
zveinn: I can make some kind of "error" here so it won't call minio.. right now it just calls minio and minio throws an error

summary-v4-object.go
zveinn: I'll also move this to msgpack

summary-v4-object.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/tinylib/msgp/msgp"
+)
+
+//go:generate msgp -unexported -file=$GOFILE
+
+// ObjectSummaryOptions ...
```
klauspost, line 33:
```suggestion
// ObjectSummaryOptions provides options for ObjectSummary call.
```

summary-v4-object.go
```diff
@@ -0,0 +1,119 @@
...
+	err = msgp.Decode(resp.Body, objectSummary)
+	if err != nil {
+		return nil, err
+	}
+
+	return
+}
+
+// ObjectMetaSummary ...
+// This struct is returned from minio when calling ObjectSummary
```
klauspost, line 77:
```suggestion
// ObjectMetaSummary is returned from minio when calling ObjectSummary.
```

summary-v4-object.go
```diff
@@ -0,0 +1,119 @@
...
+	Drive          string
+	Size           int64
+	Errors         []string
+	IsDeleteMarker bool
+	ModTime        int64
+	Signature      [4]byte
+}
+
+// ObjectPartSummary ...
+// This struct is returned from minio when calling ObjectSummary.
```
klauspost, line 92:
```suggestion
// ObjectPartSummary is returned from minio when calling ObjectSummary.
```

klauspost (approved):
Just a few cosmetics

harshavardhana (approved):

PR #354, opened 2025-04-02
Make replication diag info as optional

shtripat:
As older versions of MinIO wont be sedning this detail and we should just avoid reading this.
Also replication diag is part of MinIO details so moved under `MinioHealthInfo`.


anjalshireesh (approved):
LGTM

dhananjaykrutika (approved):

PR #355, opened 2025-04-03
Update manifest file format for batch catalog

donatello:
Need to fix this - though it is breaking it is ok as the feature is unreleased.


donatello:
PTAL @harshavardhana @krisis

donatello:
PTAL @harshavardhana @krisis @vadmeste @poornas

krisis (approved):

poornas (approved):

PR #356, opened 2025-04-05
Replication metrics

shtripat:


harshavardhana:
Please resolve the conflicts @shtripat ?

cluster-commands.go
```diff
@@ -1247,6 +1247,61 @@ func (adm *AdminClient) SiteReplicationResyncOp(ctx context.Context, site PeerIn
...
+
+// EndpointServerPools - list of list of endpoints
+type EndpointServerPools []PoolEndpoints
+
+// Endpoints - list of same type of endpoint.
+type Endpoints []Endpoint
+
+// Endpoint - any type of endpoint.
+type Endpoint struct {
+	*url.URL
```
klauspost, line 1270: How is that actually transported across the wire?

I would prefer a base type - presumably just a string.

shtripat: Will change to `string` as makes sense.

shtripat: Kept `url.URL` as non pointer as receiving side we need other fields for anonymizing the URL

user-commands.go
```diff
@@ -988,3 +988,40 @@ func (adm *AdminClient) InfoAccessKey(ctx context.Context, accessKey string) (In
 	}
 	return infoResp, nil
 }
+
+type CheckObjectManagePermissionsOpts struct {
```
harshavardhana, line 992: Add comments about each struct

harshavardhana (commented):
some missing documentation for posterity

cluster-commands.go
klauspost: How is that sent in JSON?

shtripat: The endpoint object marshalled looks as below
```
{
    "Scheme" : "http",
    "Opaque" : "",
    "User" : null,
    "Host" : "127.0.0.1:9011",
    "Path" : "/tmp/xl1/16/disk1",
    "RawPath" : "",
    "OmitHost" : false,
    "ForceQuery" : false,
    "RawQuery" : "",
    "Fragment" : "",
    "RawFragment" : "",
    "isLocal" : false,
    "poolIdx" : 1,
     "setIdx" : 2,
     "diskIdx" : 5
}
```
the ones with camelCase are top level fields where I added `json:` tags but rest from url.URL and struct attribute names are used as is.

cluster-commands.go
```diff
@@ -1247,6 +1247,62 @@ func (adm *AdminClient) SiteReplicationResyncOp(ctx context.Context, site PeerIn
...
 }
 
+// PoolEndpoints represent endpoints in a given pool
+// along with its setCount and setDriveCount.
+type PoolEndpoints struct {
+	// indicates if endpoints are provided in non-ellipses style
+	Legacy       bool      `json:"legacy"`
+	SetCount     int       `json:"setCount"`
+	DrivesPerSet int       `json:"drivesPerSet"`
+	Endpoints    Endpoints `json:"endpoints"`
```
klauspost, line 1257:
```suggestion
	Drives    Drives `json:"drives"`
```
... that's what it is, correct?

cluster-commands.go
```diff
@@ -1247,6 +1247,62 @@ func (adm *AdminClient) SiteReplicationResyncOp(ctx context.Context, site PeerIn
...
+
+// EndpointServerPools - list of list of endpoints
+type EndpointServerPools []PoolEndpoints
+
+// Endpoints - list of same type of endpoint.
+type Endpoints []Endpoint
+
+// Endpoint - any type of endpoint.
+type Endpoint struct {
+	url.URL
```
klauspost, line 1270: Maybe instead...

```suggestion
	Endpoint string `json:"endpoint"`
	Host string `json:"host"`	
```

To have the host and drive URL.

klauspost (commented):
That makes sense to me, at least ;)

cluster-commands.go
shtripat: yes, will change

cluster-commands.go
klauspost: @shtripat Then at least add it as a separate entity `UrlParts *url.URL json:"urlParts,omitempty"`

cluster-commands.go
```diff
@@ -1247,6 +1247,62 @@ func (adm *AdminClient) SiteReplicationResyncOp(ctx context.Context, site PeerIn
...
+	// indicates if endpoints are provided in non-ellipses style
+	Legacy       bool   `json:"legacy,omitempty"`
+	SetCount     int    `json:"setCount,omitempty"`
+	DrivesPerSet int    `json:"drivesPerSet,omitempty"`
+	Drives       Drives `json:"drives,omitempty"`
+	CmdLine      string `json:"cmdLine,omitempty"`
+	Platform     string `json:"platform,omitempty"`
+}
+
+// EndpointServerPools - list of list of endpoints
```
poornas, line 1262:
```suggestion
// EndpointServerPools - list of pool endpoints
```

poornas (changes requested):

klauspost (approved):
<3

poornas (approved):

shtripat:
Closing in favor of https://github.com/minio/madmin-go/pull/375 as scope if different for capturing replication diag data.

PR #357, opened 2025-04-08
ConsoleMsg is not used by open source or AIStor, mark deprecated

marktheunissen:
Explanation here: https://github.com/minio/mc/pull/5186


zveinn (approved):

harshavardhana (changes requested):
Can't remove fields just add deprecated in front and stop using it. Remove it properly in v4

PR #363, opened 2025-04-11
Upgrade github.com/shirou/gopsutil to v4

Juneezee:
v4 is the latest version with bug fixes and enhancements, notably the removal of CGO implementations on Darwin in [v4.24.9](https://github.com/shirou/gopsutil/releases/tag/v4.24.9).

Note: Due to the library's versioning policy, [v4.24.5](https://github.com/shirou/gopsutil/releases/tag/v4.24.5) is the first v4 release. It introduces a few breaking changes, but they are straightforward to adapt.


klauspost:
This would be breaking API. @harshavardhana should we do this for madmin v4?

harshavardhana (approved):

PR #368, opened 2025-04-15
Adding a slice for unknown files

zveinn:


summary-object.go
```diff
@@ -100,6 +100,18 @@ type ObjectPartSummary struct {
...
 
+// ObjectUnknownSummary is returned from minio when calling ObjectSummary.
+// This struct contains information about files that are not part of any object structure.
+type ObjectUnknownSummary struct {
+	Pool     int
+	Host     string
+	Set      int
+	Drive    string
+	Filename string
+	Size     int64
```
klauspost, line 111:
```suggestion
	ModTime time.Time
	Size     int64
```
I guess we already have it?

klauspost (commented):
Need `go generate`. Otherwise LGTM.

summary-object.go
zveinn: mod time is on the meta summary object, unless you want OS mode time for each part ?

klauspost: Mtime (File, not metadata) is actually quite useful, so see last write time was consistent. For abandoned files I can tell us when it was dumped there.

harshavardhana (approved):

PR #370, opened 2025-04-16
introduce Event type and subsystem for event recorder

Praveenrajmani:


api-events.go
```diff
@@ -0,0 +1,72 @@
...
+		resp, err := adm.executeMethod(ctx, http.MethodGet, reqData)
+		if err != nil {
+			closeResponse(resp)
+			return
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			return
+		}
+		dec := json.NewDecoder(resp.Body)
```
klauspost, line 55: If this is to be a high throughput API, consider using `msgp` for the transfer.

You will need to get rid of the `interface{}` values. Let me know if you need assistance in setting that up.

(it should be pretty similar to this)

harshavardhana:
Please rebase this with main branch now and move this code to v4

api-events.go
Praveenrajmani: these jwt libraries have claims defined in `{}interface` @klauspost let me try to simplify it

Praveenrajmani:
moving this PR to main

Praveenrajmani:
`ec` is not yet updated with `v4`, re-opening this PR

Praveenrajmani:
ec is updated with v4 now, closing this PR

PR #371, opened 2025-04-16
Add Kubernetes subsystem config constant

pjuarezd:


donatello:
@pjuarezd could you also bump up go here - https://github.com/minio/madmin-go/blob/main/go.mod#L3 ? looks like we need it to fix the Cve

donatello (approved):

ramondeklein (approved):

PR #372, opened 2025-04-17
Add Kubernetes subsystem config constant to v3

pjuarezd:
# Add Kubernetes subsystem Config constant to v3 package

eos is still using the v3 version of this package, although this commit is already merged into main branch (v4), upgrading eos to v4 might be a little too soon.

By cherry-picking [this commit](https://github.com/minio/madmin-go/commit/6e378b77df7ec72575397f35b5fc6f9cca5a54b6) to `v3` branch we can update eos to have the `kubernetes` subsystem constant with minimal update of this package and it's dependencies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/minio/madmin-go/v3
 
 go 1.24.0
 
+toolchain go1.24.2
+
 require (
 	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/dustin/go-humanize v1.0.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/minio/madmin-go/v3
 
-go 1.23.0
-
-toolchain go1.24.2
```
donatello, line 5: We need the toolchain line @pjuarezd

donatello (changes requested):

go.mod
pjuarezd: added

donatello (approved):

PR #375, opened 2025-04-21
Replication diag v2 structs

shtripat:


replication.go
```diff
@@ -25,6 +25,52 @@ import (
...
+
+// ReplDiagSiteV2 represents the replication site information
+type ReplDiagSiteV2 struct {
+	Addr         string `json:"addr,omitempty"`
+	DeploymentID string `json:"deployment_id"`
+	Online       bool   `json:"online,omitempty"`
+}
+
+// ReplDiagReplBucketV2 represents the replication target information for a bucket
+type ReplDiagReplBucketV2 struct {
```
poornas, line 57:
```suggestion
type ReplDiagBucketV2 struct {
```

replication.go
```diff
@@ -25,6 +25,52 @@ import (
...
+	Name               string                          `json:"name,omitempty"`
+	MetadataMismatches map[string]SRBucketStatsSummary `json:"metadata_mismatches,omitempty"`
+	BucketReplTargets  []BucketReplTargetV2            `json:"bucket_repl_targets,omitempty"`
```
poornas, line 60:
```suggestion
	Targets  []BucketReplTargetV2            `json:"targets,omitempty"`
```

poornas (changes requested):

health.go
```diff
@@ -1088,9 +1088,9 @@ type TLSCert struct {
 type MinioHealthInfo struct {
 	Error string `json:"error,omitempty"`
 
-	Config      MinioConfig   `json:"config,omitempty"`
-	Info        MinioInfo     `json:"info,omitempty"`
-	Replication *ReplDiagInfo `json:"replication,omitempty"`
+	Config      MinioConfig     `json:"config,omitempty"`
+	Info        MinioInfo       `json:"info,omitempty"`
+	Replication *ReplDiagInfoV2 `json:"replication,omitempty"`
```
klauspost, line 1093: Is it breaking, if so it should be separate. If not, there is no point in making a `ReplDiagInfoV2` struct.

If breaking... do

```suggestion
	Replication *ReplDiagInfo `json:"replication,omitempty"`
	ReplicationV2 *ReplDiagInfoV2 `json:"replicationV2,omitempty"`
```

If not

```suggestion
	Replication *ReplDiagInfo `json:"replication,omitempty"`
```
.. and add new fields to the existing struct.

klauspost (changes requested):

health.go
shtripat: Actually `eos` side changes not yet merged, but I understand if with merged changes in `madmin-go` released we should add new ones.

klauspost: >  but I understand if with merged changes in madmin-go released we should add new ones.

Sorry. I can't parse what you mean by that.

* If your changes are *incompatible*, make it a new JSON field and allow to un-marshal either.
* If your changes are *compatible*, just add the new stuff to existing. Drop the V2 structs.

It is one of the above. You should chose.

shtripat: Changes are actually incompatible but old `ReplDiagInfo` not yet used in `eos` code. In this case is it like, retain the name as `ReplDiagInfo` only change the attributes of the struct?

klauspost: Yeah, let's just drop the v2 then  and use this as v1👍🏼

klauspost (approved):
lgtm.

klauspost:
Does this replace https://github.com/minio/madmin-go/pull/356

shtripat:
> Does this replace #356

Yes, I will close that.

replication.go
```diff
@@ -21,124 +21,48 @@ package madmin
...
+	TotalPolicies          int              `json:"total_policies,omitempty"`
+	SyncPendingPolicies    int              `json:"sync_pending_policies,omitempty"`
+	TotalILMExpRules       int              `json:"total_ilm_exp_rules,omitempty"`
+	SyncPendingILMExpRules int              `json:"sync_pending_ilm_exp_rules,omitempty"`
+	TotalBuckets           int              `json:"total_buckets,omitempty"`
+	SyncPendingBuckets     int              `json:"sync_pending_buckets,omitempty"`
+	Errors                 Counter          `json:"errors,omitempty"`
+	Retries                Counter          `json:"retries,omitempty"`
+	Sites                  []ReplDiagSite   `json:"sites,omitempty"`
+	ReplBuckets            []ReplDiagBucket `json:"repl_buckets,omitempty"`
```
poornas, line 44: Can you add SyncUserPolicy and SyncGroupPolicy

poornas (approved):

PR #379, opened 2025-04-30
kv: Fix the config value sanitization

vadmeste:
Use strconv.Unquote() instead of reinventing the wheel.

It will detect double and single quotes and remove them. It can also skip escaped quotes.

### DIFF
--- a/parse-config.go
+++ b/parse-config.go
@@ -22,6 +22,7 @@ package madmin
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"unicode"
 
@@ -202,8 +203,11 @@ const (
 
 // SanitizeValue - this function is needed, to trim off single or double quotes, creeping into the values.
 func SanitizeValue(v string) string {
-	v = strings.TrimSuffix(strings.TrimPrefix(strings.TrimSpace(v), KvDoubleQuote), KvDoubleQuote)
-	return strings.TrimSuffix(strings.TrimPrefix(v, KvSingleQuote), KvSingleQuote)
+	unquoted, err := strconv.Unquote(v)
+	if err != nil {
+		return v
+	}
+	return unquoted
 }
 
 // EnvOverride contains the name of the environment variable and its value.


parse-config.go
```diff
@@ -202,8 +203,11 @@ const (
 
 // SanitizeValue - this function is needed, to trim off single or double quotes, creeping into the values.
 func SanitizeValue(v string) string {
-	v = strings.TrimSuffix(strings.TrimPrefix(strings.TrimSpace(v), KvDoubleQuote), KvDoubleQuote)
-	return strings.TrimSuffix(strings.TrimPrefix(v, KvSingleQuote), KvSingleQuote)
+	unquoted, err := strconv.Unquote(v)
+	if err != nil {
+		return v
+	}
```
klauspost, line 209: On second thought can we do something if it fails?  I am a bit worried that it can be abused?

Or is it not "sanitize" in that way?

vadmeste: SanitizeValue() seems to be only used in MinIO server, where it sanitizes values of the configuration. It seems to me the strconv.Unquote is very close to the existing function. 

However, I think it makes sense to close this PR and fix MinIO server instead to use strconv.Unquote instead

harshavardhana: It was done on purpose that we only handle basically cases like this

```
"value" == value
```

```
'value' == value
```

```
"'value'" == value
```

harshavardhana (changes requested):

parse-config.go
vadmeste: > "'value'" == value

This one is weird, are we sure we want to support this ?

What I wanted to fix is this:

```
mc admin config set myminio \
    notify_postgres:primary \
    format=access  \
    table="minio_events" \
    queue_dir=/tmp/minio_events \
    queue_limit=0 \
    max_open_connections=2 \
    connection_string="host=HOST dbname=DB user=MINIO password=PASSWORD port=PORT options='-c search_path=minio'" 
```

The command above does not working because it does not sanitize correctly.

The existing code converts the following:

 `"host=HOST dbname=DB user=MINIO password=PASSWORD port=PORT options='-c search_path=minio'"` 

to
 
`host=HOST dbname=DB user=MINIO password=PASSWORD port=PORT options='-c search_path=minio`

Notice the single quote after the conversion, the single quote does not have a matching one, causing later errors.

harshavardhana: yeah but this change we have to make sure is handled properly because Console UI relies on these things.

vadmeste: it is not used anymore, as far as I looked, only MinIO server uses it right now

vadmeste: I really do not think it will break anything, I doubt this requirement is used anywhere:  `"'value'" == value`

klauspost: I could see that come from script setups, replacements, etc.

Unless we are fixing a specific issue I don't see any reason to potentially break those.

PR #380, opened 2025-05-01
Adding NodeInfo and NodeList APIs

zveinn:
Waiting on: https://github.com/minio/pkg/pull/160

### DIFF
--- a/info-v4-commands.go
+++ b/info-v4-commands.go
@@ -266,7 +266,7 @@ func (adm *AdminClient) SetInfo(ctx context.Context, poolIndex int, setIndex int
 //msgp:ignore DriveInfoOpts
 type DriveInfoOpts struct{}
 
-// DriveInfo returns pool information about a specific pool referenced by poolIndex
+// DriveInfo returns information for a specific drive
 func (adm *AdminClient) DriveInfo(ctx context.Context, poolIndex, setIndex, diskIndex int, options ...func(*DriveInfoOpts)) (Disk, error) {
 	srvOpts := &DriveInfoOpts{}
 
@@ -307,6 +307,7 @@ type NodeListOpts struct {
 	Offset int
 }
 
+// NodeListResponse includes a paginated list of nodes and the total node count
 type NodeListResponse struct {
 	Nodes []Node `msg:"nodes,omitempty"`
 	Total int    `msg:"total,omitempty"`
--- a/info-v4-commands.go
+++ b/info-v4-commands.go
@@ -298,12 +298,11 @@ func (adm *AdminClient) DriveInfo(ctx context.Context, poolIndex, setIndex, disk
 	return disk, nil
 }
 
-// NodeInfoOpts ask for additional data from the server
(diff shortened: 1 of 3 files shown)


info-v4-commands.go
```diff
@@ -290,3 +297,96 @@ func (adm *AdminClient) DriveInfo(ctx context.Context, poolIndex, setIndex, disk
...
+	Offset int
+}
+
+type NodeListResponse struct {
+	Nodes []Node `msg:"nodes,omitempty"`
+	Total int    `msg:"total,omitempty"`
+}
+
+// NodeList - list all nodes in the cluster
+func (adm *AdminClient) NodeList(ctx context.Context, options ...func(*NodeListOpts)) (nodeList *NodeListResponse, err error) {
```
klauspost, line 316: Maybe just...
```suggestion
func (adm *AdminClient) NodeList(ctx context.Context, options NodeListOpts) (nodeList *NodeListResponse, err error) {
```

.. or if you do prefer the functional approach...

```suggestion
func (adm *AdminClient) NodeList(ctx context.Context, options ...func(*NodeListOpts) error) (nodeList *NodeListResponse, err error) {
```

(and then also add the funcs to set limit/offset - though tbh it all seems kinda overkill)

info-v4-commands.go
```diff
@@ -77,6 +78,12 @@ type PoolInfo struct {
 	Hosts []string `msg:"hosts,omitempty"`
 }
 
+// Node provides information on a specific node
+type Node struct {
+	Host   string `msg:"host,omitempty"`
+	Drives int    `msg:"drives,omitempty"`
```
klauspost, line 84: We don't want this to be an array of objects, so more can be added?

info-v4-commands.go
zveinn: I just made this inline with what the other functions are doing. I don't mind either way tbh

zveinn: I think the ...func stuff was to make things more extendable without breaking consumers down the line

info-v4-commands.go
zveinn: there is a NodeListResponse for that, which includes some extra stuff for pagination

info-v4-commands.go
klauspost: Yeah - structs is usually fine and non-breaking... But if you want to do the func, it should be able to return an error.

info-v4-commands.go
klauspost: `Nodes []Node` is fine... But I am thinking of the drives - or is that not what we are interested in?

harshavardhana (approved):

info-v4-commands.go
harshavardhana: are we just calling this `Drives` not `TotalDrives` or something?

zveinn: renamed

harshavardhana (approved):

PR #383, opened 2025-05-05
feat: introduce event recorder types and implementation

Praveenrajmani:


Praveenrajmani:
ec is not yet updated with v4, closing this PR for now

PR #384, opened 2025-05-06
feat: introduce event recorder types and implementation

Praveenrajmani:
Adds a new “event” recorder subsystem and implements end-to-end support for recording, serializing, and streaming error and API events.

 - Introduce EventRecorderSubSys constant and register it in subsystem sets.
 - Define event package types (Error, Audit, API, etc.) with Msgp codegen and tests.
 - Add GetErrorEvents and GetAPIEvents methods in the madmin client plus example programs.


event/api.go
```diff
@@ -0,0 +1,83 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package event
+
+import "time"
+
+//go:generate msgp $GOFILE
+
+//msgp:replace Origin with:string
```
klauspost, line 24: Is that needed? It should be able to pick it up as long as the type def is in the same file.

klauspost (approved):
Looks great!

PR #386, opened 2025-05-14
Add Excessive Prefix list to scanner metrics

klauspost:
Scanner will keep an in-memory list of these.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -267,6 +267,10 @@ type ScannerMetrics struct {
 
 	// Currently active path(s) being scanned.
 	ActivePaths []string `json:"active,omitempty"`
+
+	// Excessive prefixes.
+	// Paths that have been marked as having excessive number of entries within the last 24 hours.
+	ExcessivePrefixes []string `json:"excessive,omitempty"`
 }
 
 // Merge other into 's'.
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -332,6 +336,23 @@ func (s *ScannerMetrics) Merge(other *ScannerMetrics) {
...
+		// Add other excessive prefixes
+		for _, prefix := range other.ExcessivePrefixes {
+			merged[prefix] = struct{}{}
+		}
+		s.ExcessivePrefixes = make([]string, 0, len(merged))
+		for prefix := range merged {
+			s.ExcessivePrefixes = append(s.ExcessivePrefixes, prefix)
+		}
+		sort.Strings(s.ExcessivePrefixes)
+	}
```
anjalshireesh, line 355: Will it be simpler if `ExcessivePrefixes` is a `StringSet` ?

anjalshireesh (approved):
LGTM, one minor question

metrics.go
klauspost: Added https://github.com/minio/minio-go/pull/2107

klauspost: Not sure it is worth including for this one.

PR #391, opened 2025-05-20
Change to capture missing replication entities

shtripat:
Now it captures missing replication entities as well as part of replication diagnostics.
Also captures mismatches for the current site only as diagnostics are captured for current site and analysis.


replication.go
```diff
@@ -46,6 +46,35 @@ type ReplDiagInfo struct {
 	GroupPolMismatches     map[string]map[string]SRGroupStatsSummary  `json:"group_policy_mismatches,omitempty"`
 }
 
+// ReplDiagInfoV2 represents the replication diagnostic information to be captured
```
klauspost, line 49: Aren't you just adding new fields? Why does this need a new type?

poornas: he changed the data type to CountWithList - instead of a V2 struct, just renaming the json field should do?

@shtripat , for the diag to be useful it needs to report which peer has the latest update on the entity if there is a mismatch - maybe something like this  
```
// CountWithList is a type that holds a count and a list of items
type CountWithList struct {
	Count int      `json:"count"`
	List  []EntityDiag `json:"list,omitempty"`
}

type EntityDiag  struct {
   Name string 
    SyncFrom []string.  // deploymentid/peer name will be present if entity missing on current site or entity present but other sites have more recent update (check UpdatedAt field)
    SyncTo  []string. // deploymentid/peer name will be present if current site is latest  and other sites are missing this entity  
}
```

health.go
```diff
@@ -1105,9 +1105,9 @@ type TLSCert struct {
 type MinioHealthInfo struct {
 	Error string `json:"error,omitempty"`
 
-	Config      MinioConfig   `json:"config,omitempty"`
-	Info        MinioInfo     `json:"info,omitempty"`
-	Replication *ReplDiagInfo `json:"replication,omitempty"`
+	Config      MinioConfig     `json:"config,omitempty"`
+	Info        MinioInfo       `json:"info,omitempty"`
+	Replication *ReplDiagInfoV2 `json:"replication,omitempty"`
```
harshavardhana, line 1110: Still making breaking changes, why are we doing that?

harshavardhana (changes requested):

health.go
shtripat: The struct `ReplDiagInfo` got field types changed so created new `ReplDiagInfoV2`. I may need to introduce new type `MinioHealthInfoV2`, right?

replication.go
shtripat: As we have changed the type of existing fields (like `SyncPendingUsers`), I feel it's cleaner to have new struct `ReplDiagInfoV2` and in `MinioHealthInfo` add new field `ReplicationInfo` as below
```
ReplicationInfo *ReplDiagInfoV2 `json:"replication_info,omitempty"`
```
Also marked the old existing field as deprecated
```
Replication     *ReplDiagInfo   `json:"replication,omitempty"` // deprecated May 2025
```
Hope this makes sense @poornas @klauspost

health.go
shtripat: Added changes as per https://github.com/minio/madmin-go/pull/391#discussion_r2121125642

replication.go
klauspost: I see it, but unless you provide conversion (v1->v2) it is extremely annoying to use.

There is really nothing clean in having to duplicate all your code that reads fields.

All your code will have to be `if v.ReplicationInfo != nil { do stuff, check fields } else if v.Replication != nil { 99% duplicated code}`.

anjalshireesh: - This struct is being added purely for the purpose of health diagnostics (replication related checks will be added in `health-analyzer` that read and analyze this data)
- Currently no such check exists
- The checks always run on the new struct (if available).

i.e. we will only have `if v.ReplicationInfo != nil { do stuff, check fields }`.

`v.Replication`, even if not nil, will never be read or analyzed. It was WIP stuff that got merged and released in one version, and will always be ignored by `health-analyzer`.

poornas: since ReplDiagInfo was never in use, just need to remove it to avoid cluttering with unused structs

shtripat: > since ReplDiagInfo was never in use, just need to remove it to avoid cluttering with unused structs

One PR in `eos` as well was merged which used this struct `madmin.ReplDiagInfo` so we need it to be retained and now use new one. Right?

health.go
```diff
@@ -1105,9 +1105,10 @@ type TLSCert struct {
 type MinioHealthInfo struct {
 	Error string `json:"error,omitempty"`
 
-	Config      MinioConfig   `json:"config,omitempty"`
-	Info        MinioInfo     `json:"info,omitempty"`
-	Replication *ReplDiagInfo `json:"replication,omitempty"`
+	Config          MinioConfig     `json:"config,omitempty"`
+	Info            MinioInfo       `json:"info,omitempty"`
+	Replication     *ReplDiagInfo   `json:"replication,omitempty"` // deprecated May 2025
```
harshavardhana:
```suggestion
	Replication     *ReplDiagInfo   `json:"replication,omitempty"` // Deprecated May 2025
```

PR #395, opened 2025-05-22
Add batch job status

allanrogerr:
Allow dependent modules to derive batch job status natively, and thus allow dependencies to classify batch jobs as failed, completed or in-progress.

Also added an `Error` string field.

https://github.com/miniohq/engineering/issues/2763

### DIFF
--- a/batch-job.go
+++ b/batch-job.go
@@ -30,14 +30,25 @@ import (
 	"time"
 )
 
-// BatchJobType type to describe batch job types
-type BatchJobType string
+type (
+	// BatchJobType describes batch jobs
+	BatchJobType string
+	// BatchJobStatusType describes batch job statuses
+	BatchJobStatusType string
+)
 
 const (
 	BatchJobReplicate BatchJobType = "replicate"
 	BatchJobKeyRotate BatchJobType = "keyrotate"
 	BatchJobExpire    BatchJobType = "expire"
 	BatchJobCatalog   BatchJobType = "catalog"
+)
+
+const (
+	BatchJobStatusCompleted  BatchJobStatusType = "completed"
+	BatchJobStatusFailed     BatchJobStatusType = "failed"
+	BatchJobStatusInProgress BatchJobStatusType = "in-progress"
+	BatchJobStatusUnknown    BatchJobStatusType = "unknown"
 )
 
 // SupportedJobTypes supported job types
@@ -194,11 +205,13 @@ const BatchJobExpireTemplate = `expire:
 
 // BatchJobResult returned by StartBatchJob
 type BatchJobResult struct {
-	ID      string        `json:"id"`
-	Type    BatchJobType  `json:"type"`
-	User    string        `json:"user,omitempty"`
-	Started time.Time     `json:"started"`
-	Elapsed time.Duration `json:"elapsed,omitempty"`
+	ID      string             `json:"id"`
+	Type    BatchJobType       `json:"type"`
+	User    string             `json:"user,omitempty"`
+	Started time.Time          `json:"started"`
+	Elapsed time.Duration      `json:"elapsed,omitempty"`
+	Status  BatchJobStatusType `json:"status,omitempty"`
+	Error   string             `json:"error,omitempty"`
 }
 
 // StartBatchJob start a new batch job, input job description is in YAML.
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -567,13 +577,20 @@ func (o *BatchJobMetrics) Merge(other *BatchJobMetrics) {
...
 	}
+	// Job
 	if o.Jobs == nil {
 		o.Jobs = make(map[string]JobMetric, len(other.Jobs))
 	}
-	// Job
 	for k, v := range other.Jobs {
 		o.Jobs[k] = v
 	}
+	// Job Type
```
zveinn, line 587: maybe remove the comment or make it more descriptive ?

zveinn (approved):

metrics.go
allanrogerr: Removed redundant comment

batch-job.go
```diff
@@ -30,8 +30,12 @@ import (
 	"time"
 )
 
-// BatchJobType type to describe batch job types
-type BatchJobType string
+type (
+	// BatchJobType describes batch jobs
+	BatchJobType string
+	// BatchJobStatusType describes batch job statuses
+	BatchJobStatusType string
```
harshavardhana, line 37:
```suggestion
	BatchJobStatus string
```

batch-job.go
```diff
@@ -40,6 +44,13 @@ const (
 	BatchJobCatalog   BatchJobType = "catalog"
 )
 
+const (
+	BatchJobStatusCompleted  BatchJobStatusType = "completed"
+	BatchJobStatusFailed     BatchJobStatusType = "failed"
+	BatchJobStatusInProgress BatchJobStatusType = "in-progress"
+	BatchJobStatusUnknown    BatchJobStatusType = "unknown"
+)
```
harshavardhana, line 52:
```suggestion
	BatchJobStatusCompleted  BatchJobStatus = "completed"
	BatchJobStatusFailed     BatchJobStatus = "failed"
	BatchJobStatusInProgress BatchJobStatus = "in-progress"
	BatchJobStatusUnknown    BatchJobStatus = "unknown"
)
```

batch-job.go
```diff
@@ -194,11 +205,12 @@ const BatchJobExpireTemplate = `expire:
...
-	Type    BatchJobType  `json:"type"`
-	User    string        `json:"user,omitempty"`
-	Started time.Time     `json:"started"`
-	Elapsed time.Duration `json:"elapsed,omitempty"`
+	ID      string             `json:"id"`
+	Type    BatchJobType       `json:"type"`
+	User    string             `json:"user,omitempty"`
+	Started time.Time          `json:"started"`
+	Elapsed time.Duration      `json:"elapsed,omitempty"`
+	Status  BatchJobStatusType `json:"status"`
```
harshavardhana, line 213:
```suggestion
	Status  BatchJobStatus `json:"status"`
```

metrics.go
```diff
@@ -481,6 +481,16 @@ type BatchJobMetrics struct {
 
 	// Jobs by ID.
 	Jobs map[string]JobMetric
+
+	// Job Status by Type
+	Status map[BatchJobType]BatchJobSummaryStatus
```
harshavardhana, line 486: Can we not add this in `JobMetric` ? Because the `JobMetric.InProgress` is where I would expect this.

Also not sure why these are `float64`

harshavardhana (dismissed):

batch-job.go
allanrogerr: BatchJobStatus is an existing type; this name would cause a conflict

metrics.go
allanrogerr: I have nested `Status` within `JobMetric`. Also removed these `float64`s.

allanrogerr:
@zveinn @harshavardhana  I've simplified the change. PTAL

batch-job.go
```diff
@@ -194,11 +205,13 @@ const BatchJobExpireTemplate = `expire:
...
+	Message *string            `json:"message,omitempty"`
```
harshavardhana, line 214:
```suggestion
	Message string            `json:"message,omitempty"`
```

Doesn't need to be a pointer to be empty.

harshavardhana: and also what is this Message supposed to be anyway a comment would help?

harshavardhana: if its Error then we can simply have the Message renamed as `Error string` we use this commonly everywhere.

harshavardhana (changes requested):

batch-job.go
allanrogerr: @vadmeste Please comment on your plans for this new field.

PR #396, opened 2025-05-24
provide helper methods to consume ARNs for notification targets

harshavardhana:


arn.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if tokens[4] == "" || tokens[5] == "" {
+		return nil, fmt.Errorf("invalid ARN %s", s)
+	}
+
+	return &ARN{
+		Type:   ServiceType(tokens[2]),
+		Region: tokens[3],
+		ID:     tokens[4],
+		Bucket: tokens[5],
```
vadmeste, line 67: this should be Resource in a madmin-go v5, right ?

vadmeste (approved):

arn_test.go
```diff
@@ -0,0 +1,224 @@
+package madmin
```
klauspost, line 1: Copyright missing? :D

klauspost (approved):

PR #400, opened 2025-06-02
Introducing Telemetry Target(s) subsystem

dhananjaykrutika:
This subsystem will be used to manage telemetry target settings.

Also:
* Move utility functions to `madmin-go`
* Extract `HTTPFilter` into a separate struct.


opentelemetry.go
```diff
@@ -134,3 +140,59 @@ func (adm AdminClient) ServiceTelemetry(ctx context.Context, opts ServiceTelemet
 		}
 	}
 }
+
+// StringToTraceType - given a comma-separated string of types, returns OR'd tracetypes counterpart
+func StringToTraceType(typeStr string) TraceType {
```
klauspost:
```suggestion
// ParseTraceType - given a comma-separated string of types, returns OR'd tracetypes counterpart
func ParseTraceType(typeStr string) TraceType {
```

opentelemetry.go
```diff
@@ -134,3 +140,59 @@ func (adm AdminClient) ServiceTelemetry(ctx context.Context, opts ServiceTelemet
...
+	}
+
+	if traceType == 0 {
+		traceType = TraceS3
+	}
+	return traceType
+}
+
+// ParseSampleRate converts a sample rate from string format to float64.
+func ParseSampleRate(s string) (float64, error) {
```
klauspost, line 166: Sent tests elsewhere.

api-events.go
```diff
@@ -0,0 +1,84 @@
...
+	Origin     event.Origin  `json:"origin,omitempty"`
+	Type       event.APIType `json:"type,omitempty"`
+}
+
+// GetAPIEvents fetches the persisted API events from MinIO
+func (adm AdminClient) GetAPIEvents(ctx context.Context, opts APIEventOpts) iter.Seq2[event.API, error] {
+	return func(yield func(event.API, error) bool) {
+		apiOpts, err := json.Marshal(opts)
+		if err != nil {
+			yield(event.API{}, err)
```
harshavardhana:
```suggestion
			yield(event.API{}, err)
			return
```

api-events.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		reqData := requestData{
+			relPath: adminAPIPrefix + "/events/api",
+			content: apiOpts,
+		}
+		resp, err := adm.executeMethod(ctx, http.MethodPost, reqData)
+		if err != nil {
+			closeResponse(resp)
+			yield(event.API{}, err)
```
harshavardhana:
```suggestion
			yield(event.API{}, err)
			return
```

resp is never set when `err != nil`

api-events.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		if resp.StatusCode != http.StatusOK {
+			closeResponse(resp)
```
harshavardhana: use httpRespToErrorResponse(resp) to capture the error and yield that and return

error-events.go
```diff
@@ -0,0 +1,81 @@
...
+			content: errOpts,
+		}
+		resp, err := adm.executeMethod(ctx, http.MethodPost, reqData)
+		if err != nil {
+			closeResponse(resp)
+			yield(event.Error{}, err)
+		}
+		if resp.StatusCode != http.StatusOK {
+			closeResponse(resp)
+			return
```
harshavardhana: Same changes apply here as well as per above review..

opentelemetry.go
```diff
@@ -134,3 +140,58 @@ func (adm AdminClient) ServiceTelemetry(ctx context.Context, opts ServiceTelemet
...
+			return 0, fmt.Errorf("invalid sample rate (%s)", s)
+		}
+		pct, err := strconv.ParseFloat(strings.TrimSpace(split[0]), 64)
+		if err != nil || pct <= 0 {
+			return 0, fmt.Errorf("invalid sample rate (%s)", s)
+		}
+		return pct / 100, nil
+	}
+	// Assume it is just a number.
+	return strconv.ParseFloat(strings.TrimSpace(s), 64)
```
harshavardhana, line 196: what would this mean?

dhananjaykrutika: A number, like 0.5, 0.01 etc.

harshavardhana: Can that be commented?

dhananjaykrutika: Will add a comment.

klauspost:
@dhananjaykrutika You should be able to just do `git merge master` to merged in upstream as a single commit. It is not a big deal, we always squash+merge.

api-events.go
Praveenrajmani: fixing this in https://github.com/minio/madmin-go/pull/402

api-events.go
Praveenrajmani: fixing this in https://github.com/minio/madmin-go/pull/402

api-events.go
Praveenrajmani: fixing this in https://github.com/minio/madmin-go/pull/402

error-events.go
Praveenrajmani: fixing this in https://github.com/minio/madmin-go/pull/402

klauspost:
I don't understand what I am looking at here. How did all the changes from unrelated PRs end up in here?

dhananjaykrutika:
> I don't understand what I am looking at here. How did all the changes from unrelated (already merged) PRs end up in here?
> 
> This makes no sense to me. How is it even mergable?

I don't have a good explanation for how it ended up in this state. I will fix this (mess). Give me some time.

dhananjaykrutika:
> I don't understand what I am looking at here. How did all the changes from unrelated (already merged) PRs end up in here?
> 
> This makes no sense to me. How is it even mergable?

Fixed. :sweat_smile:

klauspost (approved):
🎉

PR #401, opened 2025-06-02
Add QoS APIs

poornas:


qos.go
```diff
@@ -0,0 +1,140 @@
...
+	Label        string                `yaml:"label,omitempty"`
+	BucketPrefix string                `yaml:"bucketPrefix"` // includes prefix if any and represents simple wildcard for bucket, prefix
+	APIRequests  []RequestLimitsConfig `yaml:"requests,omitempty" json:"requests,omitempty" mapstructure:"requests"`
+}
+
+// APILimit - define API Limit
+type APILimit struct {
+	Rate  int64  `yaml:"rate" json:"rate" mapstructure:"rate"`
+	Burst int64  `yaml:"burst" json:"burst" mapstructure:"burst"`
+	Limit string `yaml:"limit" json:"type" mapstructure:"limit"` // "concurrency" or "request" limiter
```
klauspost: Thanks for adding the `concurrency`. <3

```suggestion
	Limit string `yaml:"limit" json:"type" mapstructure:"limit"` // "concurrent" or "rps" limiter
```

I feel like those make it slightly more clear what each it doing.

qos.go
```diff
@@ -0,0 +1,140 @@
...
+type QOSConfig struct {
+	Version string   `yaml:"version"` // Version of the QoS configuration format
+	Rules   QOSRules `yaml:"rules"`
+}
+
+// QOSRules represents QoS rules for buckets
+type QOSRules struct {
+	Buckets []QOSBucketRule `yaml:"buckets"`
+	// Add new rules here
+}
```
klauspost: I don't know if this is to get a specific YAML structure, but from a Go perspective I would reduce this by one level.
```suggestion
type QOSConfig struct {
	Version string   `yaml:"version"` // Version of the QoS configuration format
	Rules   []QOSRule `yaml:"rules"`
}
```

qos.go
```diff
@@ -0,0 +1,140 @@
...
+
+// QOSBucketRule represents a single QoS rule at the bucket level.
+type QOSBucketRule struct {
+	ID           string                `yaml:"id"`
+	Label        string                `yaml:"label,omitempty"`
+	BucketPrefix string                `yaml:"bucketPrefix"` // includes prefix if any and represents simple wildcard for bucket, prefix
```
klauspost: It seems to me that adding several buckets to a single rule would remove a lot of duplicate rules if applying to many buckets.
```suggestion
	BucketPrefixes []string                `yaml:"bucketPrefixes"` // include buckets with any of these prefixes. If none, include all.
```

We can add `UserAgent []string` or other conditions later. These conditions will be AND'ed, therefore empty should include all.

qos.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// QOSRules represents QoS rules for buckets
+type QOSRules struct {
+	Buckets []QOSBucketRule `yaml:"buckets"`
+	// Add new rules here
+}
+
+// QOSBucketRule represents a single QoS rule at the bucket level.
+type QOSBucketRule struct {
```
klauspost:
```suggestion
// QOSRule represents a single QoS rule applying to one or more buckets.
type QOSRule struct {
```

qos.go
```diff
@@ -0,0 +1,140 @@
...
+	ID           string                `yaml:"id"`
+	Label        string                `yaml:"label,omitempty"`
+	BucketPrefix string                `yaml:"bucketPrefix"` // includes prefix if any and represents simple wildcard for bucket, prefix
+	APIRequests  []RequestLimitsConfig `yaml:"requests,omitempty" json:"requests,omitempty" mapstructure:"requests"`
+}
+
+// APILimit - define API Limit
+type APILimit struct {
+	Rate  int64  `yaml:"rate" json:"rate" mapstructure:"rate"`
+	Burst int64  `yaml:"burst" json:"burst" mapstructure:"burst"`
```
klauspost: A) Maybe doc what the timescales are for each.
B) I don't see "Burst" making sense with "concurrent". Maybe document that it doesn't do anything.
C) Maybe add `action: <reject|block>` that means that calls are either rejected or blocked.

qos.go
```diff
@@ -0,0 +1,140 @@
...
+	Limit string `yaml:"limit" json:"type" mapstructure:"limit"` // "concurrency" or "request" limiter
+}
+
+// RequestLimitsConfig - define Request Limits
+type RequestLimitsConfig struct {
+	API      string `yaml:"api" json:"api" mapstructure:"api"`
```
klauspost: Maybe allow multiple APIs? Or is this a `map[string]APILimit` internally?
```suggestion
	APIs      []string `yaml:"api" json:"api" mapstructure:"api"`
```

Then maybe add APIs to `APILimit` - and maybe allow prefixes?

qos.go
poornas: burst can still be useful in concurrency rate limiting for temporary spikes. Is there a need for explicit action - seems like it would be better to be a implicit server side choice

qos.go
poornas: it is a map of API's -> APILimit

qos.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return &qosCfg, nil
+}
+
+// SetQOSConfig sets the QoS configuration for the MinIO server.
+func (adm *AdminClient) SetQOSConfig(ctx context.Context, qosCfg QOSConfig) error {
+	data, err := yaml.Marshal(qosCfg)
+	if err != nil {
+		fmt.Println("Error marshalling QoS configuration:", err, string(data))
```
vadmeste: not needed

qos.go
```diff
@@ -0,0 +1,126 @@
...
+const QOSConfigVersionCurrent = "v1"
+
+// QOSConfig represents the QoS configuration with a list of rules.
+type QOSConfig struct {
+	Version string    `yaml:"version"` // Version of the QoS configuration format
+	Rules   []QOSRule `yaml:"rules"`
+}
+
+// QOSRule represents a single QoS rule at the bucket prefix or bucket wildcard level.
+type QOSRule struct {
```
vadmeste, line 39: I am guessing in the future we will add 'Condition'  inside QSRule ? like requests credentials, or anonymous requests or IP address.. but it looks like Condition will fit fine

qos.go
```diff
@@ -0,0 +1,126 @@
...
+	ID           string              `yaml:"id"`
+	Label        string              `yaml:"label,omitempty"`
+	BucketPrefix string              `yaml:"bucketPrefix"` // includes prefix if any and represents simple wildcard for bucket, prefix
+	Requests     map[string]APILimit `yaml:"requests,omitempty" json:"requests,omitempty" mapstructure:"requests"`
+}
+
+// APILimit - define API Limit
+type APILimit struct {
+	Rate  int64  `yaml:"rate" json:"rate" mapstructure:"rate"`    // requests per second for (rps), concurrent requests for (concurrency)
+	Burst int64  `yaml:"burst" json:"burst" mapstructure:"burst"` // burst size for the limiter
```
vadmeste: can you explain more what is Burst ?

qos.go
```diff
@@ -0,0 +1,126 @@
...
+
+	if err = yaml.Unmarshal(b, &qosCfg); err != nil {
+		return nil, err
+	}
+
+	return &qosCfg, nil
+}
+
+// SetQOSConfig sets the QoS configuration for the MinIO server.
+func (adm *AdminClient) SetQOSConfig(ctx context.Context, qosCfg QOSConfig) error {
```
vadmeste, line 95: I think somehow the server still needs to save a QOS config for each bucket separately, so you can load a QOS rule on the fly if it is not loaded yet in the server

Just thninking out load.. we still do not have to pass a bucket name right here.. but the server needs to separate rules per bucket in my opinion

qos.go
klauspost: > burst can still be useful in concurrency rate limiting for temporary spikes. 

Since concurrency isn't time-windowed bursts doesn't really make sense. What does the burst setting do in concurrent mode?

>  Is there a need for explicit action - seems like it would be better to be a implicit server side choice

Whether the calls are rejected or blocked will make a big difference to some clients, so I could imagine some would like it to block, but with a timeout before being rejected. We can add it later, though.

poornas: shelving action: block/reject for later

it seems like setting an arbitrary number for concurrency would be harder from user perspective - a burst can provide some leeway to handle spikes at a point in time in case the concurrency setting is misestimated

qos.go
poornas: if additional rules are added - we can add a Type field to group relevant rules as needed

qos.go
poornas: burst accomodates for spikes in requests received - e.g if rate limiter uses rate of 1rq/s vs 1rq/s with burst of 5rq/s. If 100 requests received in that one second, the first rate limiter setting denies 99 requests whereas the second will allow additional 5 requests in that window

qos.go
poornas: hmm, rate limiting settings set at bucket level will not give an overall picture of whether sane rules are in place for the server. 

The QOS rule specified is per node and can be considered a list of applicable rule template (either as bucket specific prefix level template or a more generic wildcard). 

Updating/Adding a rule on this config should apply the change to API throttles for each relevant bucket that matches this wildcard.  basically when a new request comes in that has rate limiting set for that API, an API throttle is created in memory for this bucket. 

config updates relevant to this API will reflect in the in-memory API throttle for each bucket it is applicable for.

poornas:
> I see sharing rate-limits for several buckets is a thing. So If your rule bucket prefix is `bucket` and you have `bucket1`, `bucket2` and `bucket3`, I would assume that the rate-limit is applied _in total_ for these buckets ,and not individually.

As mentioned above, the qos rule is actually a template - if a rule specifies rate limit  of 100rq/s for buckets with pattern `bucket*`,  each of `bucket1`, `bucket2` and `bucket3` will be given 100rq/s per node before rate limiter rejects the requests

> Similar, assuming there are several matching rules for a bucket - are they all applied at the same time? I would kinda assume they are.

> So a rule with prefix `""` (all buckets) can have a rate limit of 100 requests and you can then do a rule for `"mybucket"` with a rate limit of 10. So the server is limited to 100/r and `"mybucket"` in particular can only have 10? Is that how it works?

the rules need to be in order of highly specific to the more generic or "*" rules at the end - the first applicable and most specific rule is picked for the API in question

klauspost:
@poornas I don't agree that what you are describing is something that makes sense for customers.

Say you can handle 300 req/s and you have 10 buckets, you would need to limit each to 30/s to ensure that you don't go over your target. That means that if one bucket is saturated you are "wasting" 270 req/s that the system can provide.

You want the limits to be as "wide" as possible to more accurately match the system. It is fine that you can set a per-bucket limit, but what we are really interested in is system-wide limits, since that is what we want to protect.

qos.go
```diff
@@ -0,0 +1,130 @@
...
+type RequestLimits struct {
+	API      string `yaml:"api" json:"api" mapstructure:"api"`
+	APILimit `yaml:",inline" mapstructure:",squash"`
+}
+
+// APILimit - define API Limit
+type APILimit struct {
+	Rate  int64  `yaml:"rate" json:"rate" mapstructure:"rate"`    // requests per second for (rps), concurrent requests for (concurrency)
+	Burst int64  `yaml:"burst" json:"burst" mapstructure:"burst"` // burst size for the limiter
+	Limit string `yaml:"limit" json:"type" mapstructure:"limit"`  // "concurrency" or "rps" limiter
```
zveinn: maybe turns this into types ? anyone that's using the api would have to read the docs for this to make sense

zveinn (commented):
Form what I remember in the presentation, each rule has to be applied per bucket ? we should probably allow cluster wide rules. We have some clients that have hundreds of buckets. 

Also, we have clients that constantly create and delete buckets.

qos.go
klauspost: Are you thinking bool types? Or enum?

harshavardhana:
> The actual design needs more discussion - this is implemented similar to minwall as a first pass, agree with you @klauspost that the design goal should be to protect server, which concurrency rate limiting is more geared for. The bucket level rate limiting is pulled off minwall and more suited for API request limiting on specific buckets.

This must be bucket centric @poornas global QoS might not be needed right now as its harder to manage things.

poornas:
> This must be bucket centric @poornas global QoS might not be needed right now as its harder to manage things.

https://github.com/miniohq/ec/pull/80#issuecomment-3120850269 - this can still be bucket centric as rules can be defined for a bucket, prefix or even at the node level with wildcards, with rule priority deciding which rule is applied. 
Rather than rate limit entire bucket regardless of API, it may be better to provide sane limits for API's which tax the system highest

poornas:
> > [miniohq/ec#80 (comment)](https://github.com/miniohq/ec/pull/80#issuecomment-3120850269) - this can still be bucket centric as rules can be defined for a bucket, prefix or even at the node level with wildcards, with rule priority deciding which rule is applied.
> > Rather than rate limit entire bucket regardless of API, it may be better to provide sane limits for API's which tax the system highest
> 
> I was thinking that this would be a bucket centric config, it is okay to have it global config as well but we have to mandate that customers must set the values per bucket - we can't take X amount and divide that by number of buckets - that kind of approach might not work.

true - the config settings must be sane and based on actual metrics and workload seen, not theoretical values

qos.go
```diff
@@ -0,0 +1,120 @@
...
+		relPath:     adminAPIPrefix + "/get-qos-config",
+		queryValues: queryValues,
+	}
+
+	// Execute GET on /minio/admin/v4/get-qos-config
+	resp, err := adm.executeMethod(ctx, http.MethodGet, reqData)
+	defer closeResponse(resp)
+	if err != nil {
+		return nil, err
+	}
```
csumisha, line 72: I think better defer after checking the error. If there is any error it might panic.

qos.go
```diff
@@ -0,0 +1,120 @@
...
+
+package madmin
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"net/url"
+
+	"gopkg.in/yaml.v2"
```
harshavardhana:
```suggestion
	"gopkg.in/yaml.v3"
```

qos.go
harshavardhana: @vadmeste what we are making sure is specifying bucket is not optional - which is how the yaml must be validated IMO. 

The QoS must be bucket centric, but how it is saved on the server and loaded is not a directly requirement on the bucket itself.

vadmeste: okay, I was hoping of not putting all eggs in one basket because if the config in the server will be stored in a single file, there is some risks that the user will need to reconfigure QoS for all buckets, unless he does mc admin qos export/import command

poornas:
Can this be reviewed again? Just to clarify, many of the original questions have been resolved
```
 mc qos rule ls sitea
QoS Configuration Rules:
API: s3.List*
  Priority     : 1 Bucket Prefix: nonprod*
  Rule ID      : restrict-non-prod-buckets
  Limits       : Rate: 30, Burst: 0, Limit: concurrency
API: s3.PutObject
  Priority     : 1 Bucket Prefix: bucket1/limitme*
  Rule ID      : limit-pfx1
  Limits       : Rate: 30, Burst: 30, Limit: rps
  Priority     : 2 Bucket Prefix: bucket1/*
  Rule ID      : catchall-bucket1
  Limits       : Rate: 100, Burst: 30, Limit: rps
API: s3.Get*
  Priority     : 1 Bucket Prefix: bucket*
  Rule ID      : f4a98a5d-d693-4136-822a-8e0b566dd91f
  Limits       : Rate: 500, Burst: 500, Limit: rps
```

what’s changed: 
1. QoS enforced by API in order of rule priority (ascending)
2. wildcard rules with partial bucket name - e.g. List rule above is no longer treated like a template. It restricts all buckets starting with nonprod to be limiting to 30 concurrent req/s for the node.
3. removed burst for concurrency limiter

qos.go
harshavardhana: you resolved this but didn't use v3 why? @poornas

qos.go
harshavardhana: Yeah, but I think having a single file is also easier to add QoS on the server as a handler. Otherwise, per-bucket QoS might be different and cannot be applied at the handler layer.

Setting and applying QoS at the object level, not at the handler level, is a possibility. I don't know what benefits or problems it might bring, but it's something to consider; there could be some advantages.

So, we can technically have QoS applied at the object level as a function() on the interface. This function is called to check if the incoming request is "throttled" or not. 

So, to truly do bucket-centric config, we need to think about the server implementation differently. It is doable. - @poornas thoughts?

harshavardhana: If we decide on bucket centric then this API must be in minio-go not here.

poornas: central QOS config gives a good picture of rate limiting rules in place - bucket centric config will obfuscate this a bit.

the QoS conf here already supports bucket centric enforcement, if rules are defined as such. We should discuss what overall objectives we want to achieve with this feature

PR #404, opened 2025-06-12
New query APIs

zveinn:
# has to be merged before: https://github.com/miniohq/eos/pull/953
# related https://github.com/miniohq/ec/pull/91

This PR introduces new APIs for API Queries that are in development


info-v4-resources.go
```diff
@@ -0,0 +1,476 @@
...
+	"context"
+	"net/http"
+	"net/url"
+	"strconv"
+
+	"github.com/tinylib/msgp/msgp"
+)
+
+//msgp:clearomitted
+//msgp:tag json
```
klauspost: This will use the `json:` for names. Remove to use the intended `msg:`

info-v4-resources.go
```diff
@@ -0,0 +1,476 @@
...
+
+// ClusterResource represents high-level information about the cluster
+type ClusterResource struct {
+	// Basics
+	Mode         string       `json:"mode" msg:"m"`
+	Domains      []string     `json:"domains,omitempty" msg:"d"`
+	Region       string       `json:"region,omitempty" msg:"r"`
+	DeploymentID string       `json:"deploymentId" msg:"di"`
+	PoolCount    int          `json:"poolCount" msg:"pc"`
+	PoolsLayout  []PoolLayout `json:"poolsLayout,omitempty" msg:"pl"`
```
```suggestion
	PoolsLayout  []PoolLayout `json:"poolsLayout,omitempty" msg:"pl,omitempty"`
```

info-v4-resources.go
```diff
@@ -0,0 +1,476 @@
...
+	OfflineDrives int `json:"offlineDrives" msg:"fd"`
+}
+
+// ServicesResourceInfo holds information about services connected to the cluster
+type ServicesResourceInfo struct {
+	LDAP          LDAP                          `json:"ldap" msg:"l"`
+	Logger        []Logger                      `json:"logger,omitempty" msg:"lg"`
+	Audit         []Audit                       `json:"audit,omitempty" msg:"a"`
+	Notifications []map[string][]TargetIDStatus `json:"notifications,omitempty" msg:"n"`
+	KMSStatus     []KMS                         `json:"kms,omitempty" msg:"k"`
```
```suggestion
	Logger        []Logger                      `json:"logger,omitempty" msg:"lg,omitempty"`
	Audit         []Audit                       `json:"audit,omitempty" msg:"a,omitempty"`
	Notifications []map[string][]TargetIDStatus `json:"notifications,omitempty" msg:"n,omitempty"`
	KMSStatus     []KMS                         `json:"kms,omitempty" msg:"k,omitempty"`
```

info-v4-resources.go
```diff
@@ -0,0 +1,476 @@
...
+// PaginatedPoolsResponse represents a paginated response for pools
+type PaginatedPoolsResponse struct {
+	Results []PoolResource `json:"results" msg:"r"`
+	Count   int            `json:"count" msg:"c"`
+}
+
+// PaginatedNodesResponse represents a paginated response for nodes
+type PaginatedNodesResponse struct {
+	Results []NodeResource `json:"results" msg:"r"`
+	Count   int            `json:"count" msg:"c"`
```
klauspost, line 46: Is `Count` the same as `len(Results)` or the total result count?

Consider adding "Offset", which reflects the current offset of the response.

zveinn: Yea, I had that discussion before but apparently the UI is relying on count so they don't have to do an "undefined" and "null" check. Javascript.. am I right ? 

Regarding offset, it's laready being sent from the client so there is no need to reflect it tbh. The only thing that I was thinking about adding is the "total" so that proper pagination could be created. Might actually just throw that in there right now.

PR #406, opened 2025-07-01
Rename redundant bucket fields in batch catalog

allanrogerr:
Currently, `mc batch status` suggests there are more buckets that can be processed e.g. 
```
  "catalog": {
   "lastBucketScanned": "dplus-qa-l2v-origin",
   "lastObjectScanned": "test/uhd_master_all.m3u8",
   "lastBucketMatched": "dplus-qa-l2v-origin",
   "lastObjectMatched": "test/uhd_master_all.m3u8",
```
See https://subnet.min.io/issues/14452?msgIndex=180

The proposed solution is to modify the struct controlling this output in `madmin-go`, then `ec` and finally `eos`


metrics.go
```diff
@@ -538,9 +538,8 @@ type KeyRotationInfo struct {
 }
 
 type CatalogInfo struct {
-	LastBucketScanned string `json:"lastBucketScanned"`
+	Bucket            string `json:"bucket"`
 	LastObjectScanned string `json:"lastObjectScanned"`
-	LastBucketMatched string `json:"lastBucketMatched"`
```
donatello, line 543: I think better to deprecate existing and add new fields to avoid a major version bump

allanrogerr: Resolved

shtripat:
This would need changes in `eos` batch catalog tests as we refer matched bucket field for verification. When we update `eos` go.mod to use latest madmin-go post this change, @allanrogerr

klauspost:
@allanrogerr Removed subnet link.

metrics.go
klauspost: @allanrogerr You need to regenerate. Your change isn't in the serializer code.

allanrogerr: @klauspost Resolved. Thank you!

klauspost: Seems like the CI check was broken: https://github.com/minio/madmin-go/pull/407

metrics.go
```diff
@@ -538,9 +538,10 @@ type KeyRotationInfo struct {
 }
 
 type CatalogInfo struct {
-	LastBucketScanned string `json:"lastBucketScanned"`
+	Bucket            string `json:"bucket"`
+	LastBucketScanned string `json:"lastBucketScanned"` // Deprecated 07/01/2025; Replaced by `Bucket`
```
```suggestion
	LastBucketScanned string `json:"lastBucketScanned,omitempty"` // Deprecated 07/01/2025; Replaced by `Bucket`
```

metrics.go
```diff
@@ -538,9 +538,10 @@ type KeyRotationInfo struct {
...
 	LastObjectScanned string `json:"lastObjectScanned"`
-	LastBucketMatched string `json:"lastBucketMatched"`
+	LastBucketMatched string `json:"lastBucketMatched"` // Deprecated 07/01/2025; Replaced by `Bucket`
```
```suggestion
	LastBucketMatched string `json:"lastBucketMatched,omitempty"` // Deprecated 07/01/2025; Replaced by `Bucket`
```

harshavardhana (approved):

PR #409, opened 2025-07-07
remove unncessary CPU freqStats

harshavardhana:
CPU stats are usually empty, capturing
this is useless for all practical purposes.

this adds some insane amount of information,
with literally no useful information to SUBNET,
trim this down.

### DIFF
--- a/cpu_linux.go
+++ b/cpu_linux.go
@@ -2,7 +2,7 @@
 // +build linux
 
 //
-// Copyright (c) 2015-2024 MinIO, Inc.
+// Copyright (c) 2015-2025 MinIO, Inc.
 //
 // This file is part of MinIO Object Storage stack
 //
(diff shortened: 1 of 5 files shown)


anjalshireesh:
@harshavardhana this contains the CPU governor information - we use it to generate warning/error if it is not set to `performance`

harshavardhana:
> @harshavardhana this contains the CPU governor information - we use it to generate warning/error if it is not set to `performance`

can you DM me where its used? - I can try to trim it down.

harshavardhana:
Okay agreed, included the governor part and rest are not needed - remove them.

cpu_linux.go
```diff
@@ -23,37 +23,33 @@
...
-			ScalingMinimumFrequency:  stat.ScalingMinimumFrequency,
-			ScalingMaximumFrequency:  stat.ScalingMaximumFrequency,
-			AvailableGovernors:       stat.AvailableGovernors,
-			Driver:                   stat.Driver,
-			Governor:                 stat.Governor,
-			RelatedCpus:              stat.RelatedCpus,
-			SetSpeed:                 stat.SetSpeed,
+		stats = append(stats, CPUFreqStats{
+			Name:     "cpu" + strconv.Itoa(i),
+			Governor: string(content),
```
anjalshireesh: Need to remove newline from the content

```json
          "freq_stats": [                                                                                                                                                                                                                                                                                             
            {                                                                                                                                                                                                                                                                                                         
              "Name": "cpu0",                                                                                                                                                                                                                                                                                         
              "Governor": "powersave\n"                                                                                                                                                                                                                                                                               
            }                                                                                                                                                                                                                                                                                                         
          ] 
```

cpu_linux.go
```diff
@@ -23,37 +23,33 @@
...
-func getCPUFreqStats() ([]CPUFreqStats, error) {
-	fs, err := sysfs.NewFS("/sys")
-	if err != nil {
-		return nil, err
-	}
+func getCPUFreqStats() (stats []CPUFreqStats, err error) {
+	for i := 0; ; i++ {
+		governorPath := filepath.Join(
+			"/sys/devices/system/cpu",
+			"cpu"+strconv.Itoa(i),
```
anjalshireesh: Might be better to store this in a variable `cpuName := "cpu"+strconv.Itoa(i)` and use in both the places.

harshavardhana: done

klauspost:
Isn't the main issue that it is duplicated for all cores? The info is good to have, but we don't need all the copies.

PR #410, opened 2025-07-07
Remove CPU frequency duplicates

klauspost:
Remove similar entries from the CPUFreqStats results.

Alternative to #409 - keeps weird values and overall settings.

### DIFF
--- a/cpu_linux.go
+++ b/cpu_linux.go
@@ -23,6 +23,9 @@
 package madmin
 
 import (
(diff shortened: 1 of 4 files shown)


harshavardhana:
It isn't just about duplicates, we don't even need the implementation used in getCPUFreqStats() - the PR #409 does this by ensuring we are only looking for "governor" settings for just one CPU.

PR #416, opened 2025-07-29
Add HealUncommittedScan heal scan mode

vadmeste:
Add an option for rapid scan feature

### DIFF
--- a/heal-commands.go
+++ b/heal-commands.go
@@ -49,7 +49,7 @@ const (
 	HealDeepScan
 
 	// HealUncommittedScan will only pick objects that are dangling (uncommitted objects in the namespace without quorum)
-	HealLowQuorumScan
+	HealUncommittedScan
 )
 
 // HealOpts - collection of options for a heal sequence


heal-commands.go
```diff
@@ -51,13 +51,16 @@ const (
...
+	Recursive bool         `json:"recursive"`
+	DryRun    bool         `json:"dryRun"`
+	Remove    bool         `json:"remove"`
+	ScanMode  HealScanMode `json:"scanMode"`
+
+	Recreate     bool `json:"recreate"`     // Rewrite all resources specified at the bucket or prefix.
+	UpdateParity bool `json:"updateParity"` // Update the parity of the existing object with a new one
+	NoLock       bool `json:"nolock"`
+
+	RapidClean bool `json:"rapidClean"`
```
harshavardhana: I think you can add it here `HealRapidScan` as ScanMode instead of RapidClean boolean which feels out of place.

vadmeste: I made HealScanMode a bitwise options.. rapid clean is only a listing mode

heal-commands.go
```diff
@@ -40,13 +40,16 @@ type HealScanMode int
...
 
 	// HealNormalScan checks if parts are present and not outdated
-	HealNormalScan
+	HealNormalScan = 1 << iota
 
 	// HealDeepScan checks for parts bitrot checksums
 	HealDeepScan
+
+	// HealLowQuorumScan will only pick objects that are obviously dangling (low quorum)
+	HealLowQuorumScan
```
harshavardhana:
```suggestion
	HealUncommittedScan
```

heal-commands.go
```diff
@@ -40,13 +40,16 @@ type HealScanMode int
...
+	HealUnknownScan HealScanMode = 0
 
 	// HealNormalScan checks if parts are present and not outdated
-	HealNormalScan
+	HealNormalScan = 1 << iota
 
 	// HealDeepScan checks for parts bitrot checksums
 	HealDeepScan
+
+	// HealLowQuorumScan will only pick objects that are obviously dangling (low quorum)
```
harshavardhana:
```suggestion
	// HealUncommittedScan will only pick objects that are dangling (uncommitted objects in the namespace without quorum)
```

PR #420, opened 2025-08-06
heal: Fix regression in scan mode definition

vadmeste:
Fix a silly recent regression that changed HealNormalScan and HealDeepScan values.


klauspost (approved):
Does this affect the latest release?

PR #422, opened 2025-08-11
refactor: use typed HealScanMode for constant

toastsandwich:
Tried to build eos with latest code
Screenshot from 2025-08-11 10-58-15

pr changes const type from untyped int to HealScanMode

### DIFF
--- a/heal-commands.go
+++ b/heal-commands.go
@@ -43,10 +43,10 @@ const (
 	HealUnknownScan HealScanMode = 0
 
 	// HealNormalScan checks if parts are present and not outdated
-	HealNormalScan = iota
+	HealNormalScan HealScanMode = iota
 
 	// HealDeepScan checks for parts bitrot checksums
-	HealDeepScan = 1 << (iota - 1)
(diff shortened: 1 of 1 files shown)


harshavardhana:
@toastsandwich you need to make relevant changes first to the miniohq/eos this is not the correct fix.

klauspost:
Seems like most of the confusion is from https://github.com/minio/madmin-go/pull/416 - so maybe @vadmeste should be involved.

vadmeste:
The PR seems to be correct, although I did not expect Golang to throw this error.. The code is a bit weird because I did not want to break the existing constants definition and bump to v5 yet.

>  Also HealUnknownScan and HealNormalScan being the same is extremely confusing.

in fact, they are not the same value, https://go.dev/play/p/j9Jr0-YRwBd.. this is also something that suprised me

I think the PR is fine

klauspost:
> The PR seems to be correct.. 

Types aren't correct?

https://go.dev/play/p/nvhmAo2rWuj

PR #423, opened 2025-08-12
Add information about the state of the drives offline tracker files

marktheunissen:
See `ec` PR # 135


heal-commands.go
```diff
@@ -363,7 +364,39 @@ const (
 	HealingReasonOfflineDisk
 )
 
-// HealingDisk contains information about
+//go:generate stringer -type=OfflineReason -output=heal-commands-offline-reason_gen.go -trimprefix=OfflineReason $GOFILE
+type OfflineReason int8
+
+const (
+	OfflineReasonInvalid OfflineReason = iota
+	OfflineReasonNodeDown
```
vadmeste, line 372: is this too late to rename it to OfflineReasonManual ? it could be the drive unmounted as well

marktheunissen: If the drive unmounted it would trigger IOError right?

`OfflineReasonNodeDown` is only set by a routine that runs on startup, before any writes/reads on the disk, so it really will be because the whole node was down.

harshavardhana: Is this resolved?

marktheunissen: @harshavardhana I had a chat to vadmeste and we talked about adding a 3rd state which I'll put up in an aistor PR, this PR can go back to draft status until then.

marktheunissen: I've added **OfflineReasonDriveDown** and **OfflineReasonTimeout**

heal-commands.go
```diff
@@ -363,7 +364,41 @@ const (
...
+	OfflineReasonIOError
+	OfflineReasonDriveDown
+	OfflineReasonTimeout
```
vadmeste:
```suggestion
	OfflineReasonDriveIOError
	OfflineReasonDriveDown
	OfflineReasonDriveTimeout
```

marktheunissen: I'll have to change them in eos too.

vadmeste (approved):

info-commands.go
```diff
@@ -525,6 +526,7 @@ type Disk struct {
 	Utilization     float64      `json:"utilization,omitempty"`
 	Metrics         *DiskMetrics `json:"metrics,omitempty"`
 	HealInfo        *HealingDisk `json:"heal_info,omitempty"`
+	OfflineInfo     *OfflineInfo `json:"offline_info,omitempty"`
```
klauspost, line 529: You couldn't know that, but I am trying to reduce the load of DiskInfo, and remove costly stats from it, since it is passed around so much.

Therefore I don't particularly like passing around an un-bounded array here, since you can get thousands of drives easily.

So I'd prefer not to add *more* to this, but don't really have a solid alternative.

TLDR; We pass around too many DiskInfos for this to not affect things.

marktheunissen: @klauspost Yeah there is a lot here, one potential solution is to have `ec` make more API requests to hydrate / populate the offline tracker information that it needs for `mc admin heal -v`, but that is for sure slower on the client side. Would you prefer that approach?

klauspost: Yeah, much more. I don't know if there is an API to get healing information, or if it just piggybacks on DiskInfo. A Fanout/in shouldn't be too expensive.

klauspost: I've removed `metrics bool` in all of the internal DiskInfo calls as well as the non-essential metrics.

marktheunissen: @klauspost Are you ok with leaving OfflineInfo as a struct member of madmin.Disk, and then we only populate it when the handler being called is for `mc admin heal -v`? It's `/minio/admin/v4/background-heal/status`. 

madmin.Disk.OfflineInfo is a pointer like the others, so needs a nil check before use, and it has `omitempty` so in JSON won't be visible if empty.

The commit that implements this is here: https://github.com/miniohq/eos/pull/1301/commits/721656f68034fe14d4117263b28bfa34b29a26bc

marktheunissen: I chatted to Klaus in DM on Slack and he seems ok with this approach

vadmeste (approved):

harshavardhana (approved):

PR #424, opened 2025-08-12
Add API metrics to realtime metrics

klauspost:
This will allow to collect and merge HTTP metrics from all nodes in a cluster.

Preparation for historical data dashboard.

Mainly to be used for mc and internally. A separate API will be added for endpoints.

### DIFF
--- a/info-commands.go
+++ b/info-commands.go
@@ -438,6 +438,7 @@ type ServerProperties struct {
 	RuntimeVersion      string            `json:"runtime_version,omitempty"`
 	GCStats             *GCStats          `json:"gc_stats,omitempty"`
 	MinioEnvVars        map[string]string `json:"minio_env_vars,omitempty"`
+	MinioEnvHash        string            `json:"minio_env_hash,omitempty"`
 	Edition             string            `json:"edition"`
 	License             *LicenseInfo      `json:"license,omitempty"`
 	IsLeader            bool              `json:"is_leader"`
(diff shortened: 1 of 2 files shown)


cesnietor:
part of: https://github.com/miniohq/aistor-console/issues/4606

metrics.go
```diff
@@ -876,3 +900,240 @@ func (m *RuntimeMetrics) Merge(other *RuntimeMetrics) {
...
+	Header         int64 `json:"header,omitempty"`         // Requests that were rejected due to header signature.
+	Invalid        int64 `json:"invalid,omitempty"`        // Requests that were rejected due to invalid request signature.
+	NotImplemented int64 `json:"notImplemented,omitempty"` // Requests that were rejected due to not implemented API.
+}
+
+// Merge other into 'a'.
+func (a *APIStats) Merge(other APIStats) {
+	if a.StartTime != nil && other.StartTime != nil {
+		if a.StartTime.Equal(*other.StartTime) {
+			a.StartTime = other.StartTime
```
vadmeste: it feels like a typo here, a.StartTime and other.StartTime are already equal at this point

klauspost: Yeah. It should be nilled at this point instead.

klauspost: Updated. PTAL.

harshavardhana (approved):

harshavardhana:
@klauspost why are you changing the heal-commands.go again? Is that expected?

PR #425, opened 2025-08-14
[persistent-logger] rename events to logs

Praveenrajmani:
this is required for [WIP - persistent logger ft](https://github.com/balamurugana/eos/pull/1) - calling it as "logs" is more appropriate than "events"

### DIFF
--- a/api-logs.go
+++ b/api-logs.go
@@ -37,7 +37,7 @@ type APILogOpts struct {
 	Node       string        `json:"node,omitempty"`
 	API        string        `json:"api,omitempty"`
 	Bucket     string        `json:"bucket,omitempty"`
-	Object     string        `json:"object,omitempty"`
+	Prefix     string        `json:"prefix,omitempty"`
 	StatusCode int           `json:"statusCode,omitempty"`
 	Interval   time.Duration `json:"interval,omitempty"`
 	Origin     log.Origin    `json:"origin,omitempty"`
--- a/log/api.go
+++ b/log/api.go
@@ -17,7 +17,15 @@
 
 package log
 
-import "time"
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+//msgp:clearomitted
+//msgp:timezone utc
+//msgp:tag json
 
 //go:generate msgp $GOFILE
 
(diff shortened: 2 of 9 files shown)


klauspost:
Really, 2 months old! ok

api-logs.go
```diff
@@ -28,46 +28,46 @@ import (
 	"net/http"
```
balamurugana, line 28: Can we add `XXHash uint64` field to each log structure in this PR?

balamurugana (approved):

Praveenrajmani:
PTAL @harshavardhana @klauspost

log/api.go
```diff
@@ -15,9 +15,12 @@
...
-package event
+package log
 
-import "time"
+import (
+	"strings"
+	"time"
+)
 
 //go:generate msgp $GOFILE
```
klauspost, line 31: Use the json tags, so omitempty will be effective. Using clearomitted if data structure should be reused. Not sure if there are any timestamps, but might as well add the tag.

```suggestion

//msgp:clearomitted
//msgp:timezone utc
//msgp:tag json

//go:generate msgp $GOFILE

```

Actually I will see if I can add it so these tags can be specified on the cmdline.

log/audit.go
```diff
@@ -15,9 +15,12 @@
...
-package event
+package log
 
-import "time"
+import (
+	"strings"
+	"time"
+)
 
 //go:generate msgp $GOFILE
```
klauspost, line 29:
```suggestion
//msgp:clearomitted
//msgp:timezone utc
//msgp:tag json

//go:generate msgp $GOFILE
```

(and regen of course)

log/error.go
```diff
@@ -15,9 +15,11 @@
...
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-package event
+package log
 
 import (
+	"fmt"
+	"strings"
 	"time"
 )
```
klauspost, line 26: Same here...

klauspost (approved):
lgtm.

We can add it to the `go:generate` soon: https://github.com/tinylib/msgp/pull/401

PR #426, opened 2025-08-15
Add custom `hash` field to hold server environment information

ramondeklein:
Add the hash values for the server's environment information.

Server hash (SHA256) should be calculated like this:
1. Sort all the `_MINIO_` and `MINIO_` environment variables on the variable name.
2. Write `<var>=<value>` to the hash (without a trailing newline).
3. Generate hash in base64 format.

### DIFF
--- a/info-commands_gen.go
+++ b/info-commands_gen.go
@@ -8752,7 +8752,7 @@ func (z *ServerProperties) DecodeMsg(dc *msgp.Reader) (err error) {
 		err = msgp.WrapError(err)
 		return
 	}
-	var zb0001Mask uint16 /* 16 bits */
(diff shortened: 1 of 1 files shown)


info-commands.go
```diff
@@ -438,6 +438,7 @@ type ServerProperties struct {
 	RuntimeVersion      string            `json:"runtime_version,omitempty"`
 	GCStats             *GCStats          `json:"gc_stats,omitempty"`
 	MinioEnvVars        map[string]string `json:"minio_env_vars,omitempty"`
+	MinioEnvHash        string            `json:"minio_env_hash,omitempty"`
```
allanrogerr, line 441: Looks like your missing running `go generate info-commands.go`

PR #427, opened 2025-08-19
New LogHost field for logger/log/entry.go and logger/audit/entry.go

toastsandwich:
This new field will give information about which node has generated the logs

### DIFF
--- a/logger/audit/entry.go
+++ b/logger/audit/entry.go
@@ -54,17 +54,18 @@ type Entry struct {
 		TimeToResponse      string          `json:"timeToResponse,omitempty"`
 		TimeToResponseInNS  string          `json:"timeToResponseInNS,omitempty"`
 	} `json:"api"`
-	RemoteHost string                 `json:"remotehost,omitempty"`
-	RequestID  string                 `json:"requestID,omitempty"`
-	UserAgent  string                 `json:"userAgent,omitempty"`
-	ReqPath    string                 `json:"requestPath,omitempty"`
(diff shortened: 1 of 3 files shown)


logger/audit/entry.go
```diff
@@ -54,17 +54,18 @@ type Entry struct {
...
-	ReqHeader  map[string]string      `json:"requestHeader,omitempty"`
-	RespHeader map[string]string      `json:"responseHeader,omitempty"`
-	Tags       map[string]interface{} `json:"tags,omitempty"`
+	LogHost    string            `json:"loghost,omitempty"`
+	RemoteHost string            `json:"remotehost,omitempty"`
+	RequestID  string            `json:"requestID,omitempty"`
+	UserAgent  string            `json:"userAgent,omitempty"`
+	ReqPath    string            `json:"requestPath,omitempty"`
+	ReqHost    string            `json:"requestHost,omitempty"`
+	ReqNode    string            `json:"requestNode,omitempty"`
```
harshavardhana, line 63: this is the logHost btw

logger/log/entry.go
```diff
@@ -54,6 +54,7 @@ type API struct {
 
 // Entry - defines fields and values of each log entry.
 type Entry struct {
+	LogHost      string         `json:"loghost,omitempty"`
```
harshavardhana, line 57: already has RemoteHost check what is that for..

zveinn: I told toasty to add it, I didn't check the ReqNode properly. I'll close this

Issue #429, opened 2025-08-22
Add method LookupEnv 


toastsandwich:
### Current func for Lookup
```go
func (c *SubsysConfig) Lookup(key string) (val string, present bool) {
	if c.kvIndexMap == nil {
		return "", false
	}

	idx, ok := c.kvIndexMap[key]
	if !ok {
		return "", false
	}
	if c.KV[idx].EnvOverride != nil {
		return c.KV[idx].EnvOverride.Value, true
	}
	return c.KV[idx].Value, true
}
```

### Change to 
```go
func (c *SubsysConfig) Lookup(key string) (val string, envVal string, present bool) {
	if c.kvIndexMap == nil {
		return
	}

	idx, ok := c.kvIndexMap[key]
	if !ok {
		return
	}
        
	if c.KV[idx].EnvOverride != nil {
		envVal = c.KV[idx].EnvOverride.Value
	}
	return c.KV[idx].Value, envVal, true
}
```

### why ?
This change will not overwrite the config value.

ps: this can be a breaking change

PR #430, opened 2025-08-22
Add method Lookupenv

toastsandwich:
# Description
PR adds new method Lookupenv. Its different from Lookup method, as it doesnt overwrite env values.
# Reference
#429

### DIFF
--- a/parse-config.go
+++ b/parse-config.go
@@ -270,6 +270,26 @@ func (c *SubsysConfig) Lookup(key string) (val string, present bool) {
 	return c.KV[idx].Value, true
 }
 
+// LookupEnv resolves the value of a config parameter.
+func (c *SubsysConfig) Lookupenv(key string) (val string, envVal string, present bool) {
+	if c.kvIndexMap == nil {
+		return
+	}
+
+	idx, ok := c.kvIndexMap[key]
+	if !ok {
+		return
+	}
+
+	if evo := c.KV[idx].EnvOverride; evo != nil {
+		envVal = evo.Value
+		present = true
+	}
+
+	val = c.KV[idx].Value
+	return val, envVal, true
+}
+
 var (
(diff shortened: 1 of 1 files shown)


klauspost:
Linter has a point ;)

parse-config.go
```diff
@@ -270,6 +270,26 @@ func (c *SubsysConfig) Lookup(key string) (val string, present bool) {
 	return c.KV[idx].Value, true
 }
 
+// Lookupenv resolves the value of a config parameter.
```
klauspost: More specifics pls... It is not really obvious from the code what it is doing (to me at least)

parse-config.go
```diff
@@ -270,6 +270,26 @@ func (c *SubsysConfig) Lookup(key string) (val string, present bool) {
...
+func (c *SubsysConfig) Lookupenv(key string) (val string, envVal string, present bool) {
```
klauspost:
```suggestion
func (c *SubsysConfig) LookupEnv(key string) (confVal, envVal string, present bool) {
```

parse-config.go
```diff
@@ -270,6 +270,26 @@ func (c *SubsysConfig) Lookup(key string) (val string, present bool) {
...
+	if c.kvIndexMap == nil {
+		return
+	}
+
+	idx, ok := c.kvIndexMap[key]
+	if !ok {
+		return
```
klauspost, line 282: So if it doesn't exist in conf, we don't care?

toastsandwich: yup

toastsandwich: we are looking for sub system configuration, if nothing is present there, we can just return false.

klauspost: Cool. document that, since it isn't obvious.

klauspost (approved):

PR #432, opened 2025-08-22
Adding sorting options to the new Query APIs

zveinn:
https://github.com/miniohq/eos/pull/1369

### DIFF
--- a/info-v4-resources.go
+++ b/info-v4-resources.go
@@ -277,6 +277,8 @@ func (adm *AdminClient) ServicesQuery(ctx context.Context, options ...func(*Serv
 //
 //msgp:ignore PoolsResourceOpts
 type PoolsResourceOpts struct {
+	// Limit defaults to 100 if set to 0.
+	// A limit of -1 will return all results.
 	Limit  int
 	Offset int
 	Filter string
@@ -295,9 +297,8 @@ func (adm *AdminClient) PoolsQuery(ctx context.Context, options *PoolsResourceOp
 	values := make(url.Values)
 
 	if options != nil {
-		if options.Limit > 0 {
-			values.Set("limit", strconv.Itoa(options.Limit))
-		}
+		values.Set("limit", strconv.Itoa(options.Limit))
+
 		if options.Offset > 0 {
 			values.Set("offset", strconv.Itoa(options.Offset))
 		}
@@ -342,6 +343,8 @@ func (adm *AdminClient) PoolsQuery(ctx context.Context, options *PoolsResourceOp
 //
 //msgp:ignore NodesResourceOpts
 type NodesResourceOpts struct {
+	// Limit defaults to 100 if set to 0.
+	// A limit of -1 will return all results.
 	Limit  int
 	Offset int
 	Filter string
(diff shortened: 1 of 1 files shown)


info-v4-resources.go
```diff
@@ -447,6 +478,14 @@ func (adm *AdminClient) ErasureSetsQuery(ctx context.Context, options *ErasureSe
...
 			values.Set("filter", options.Filter)
 		}
+		if options.Sort != "" {
+			values.Set("sort", options.Sort)
+		}
+		if options.SortReversed {
+			values.Set("sortReversed", "true")
+		} else {
+			values.Set("sortReversed", "false")
+		}
```
harshavardhana, line 520: don't you need `opts.Sort != ""` when setting `sortReversed` ?

info-v4-resources.go
```diff
@@ -34,34 +36,42 @@ import (
...
 type PaginatedErasureSetsResponse struct {
-	Results []ErasureSetResource `json:"results" msg:"r,omitempty"`
-	Count   int                  `json:"count" msg:"c"`
-	Total   int                  `json:"total" msg:"t"`
-	Offset  int                  `json:"offset" msg:"o"`
+	Results      []ErasureSetResource `json:"results" msg:"r,omitempty"`
+	Count        int                  `json:"count" msg:"c"`
+	Total        int                  `json:"total" msg:"t"`
+	Offset       int                  `json:"offset" msg:"o"`
+	Sort         string               `json:"sort" msg:"s"`
```
harshavardhana, line 74: document what all values `Sort` might contain, looks like its a free-form string if it has a defined type then please define them as constants

info-v4-resources.go
```diff
@@ -471,3 +510,46 @@ func (adm *AdminClient) ErasureSetsQuery(ctx context.Context, options *ErasureSe
...
+			if reversed {
+				return fieldI.String() > fieldJ.String()
+			}
+			return fieldI.String() < fieldJ.String()
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			if reversed {
+				return fieldI.Int() > fieldJ.Int()
+			}
+			return fieldI.Int() < fieldJ.Int()
+		case reflect.Float32, reflect.Float64:
```
```suggestion
		case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
			if reversed {
				return fieldI.Uint() > fieldJ.Uint()
			}
			return fieldI.Uint() < fieldJ.Uint()
		case reflect.Float32, reflect.Float64:
```

info-v4-resources.go
klauspost: AFAICT calling SortSlice with `""` field will just make it return.

info-v4-resources.go
klauspost: Yeah, documentation is needed.

info-v4-resources.go
zveinn: Yes, I don't want to sort as a default since we might have thousands of elements. So if you want to sort or reverse sort, you have to specify which field. Otherwise no sorting happens. I'll add that to the docs.

zveinn:
@klauspost the new sorting solution with additional vibed tests has been added. 
@harshavardhana documentation for the SortReversed and Sort fields has been updated to specify what is being sorted and what types are supported.

PR #443, opened 2025-09-13
Add support of service rolling restart/update

vadmeste:
- Cancel restart to cancel a long rolling restart
- BumpAPI is a new API to ask the cluster to use a newer API/backend format


update-commands.go
```diff
@@ -81,3 +85,35 @@ func (adm *AdminClient) ServerUpdate(ctx context.Context, opts ServerUpdateOpts)
...
+	Offline bool   `json:"offline"`
+	Error   string `json:"error,omitempty"`
+}
+
+type ClusterBumpAPIResp struct {
+	Nodes map[string]NodeBumpAPIResp `json:"nodes,omitempty"`
+	Error string                     `json:"error,omitempty"`
+}
+
+func (adm *AdminClient) BumpAPI(ctx context.Context, dryRun bool) (r ClusterBumpAPIResp, err error) {
```
harshavardhana: what is the purpose of this API can you describe it?

vadmeste:
I added some comments @harshavardhana. The PR will allow the user to use the new backend format + node API version. It is just a bump version without any other parameter that the user has to specify

update-commands.go
```diff
@@ -81,3 +87,73 @@ func (adm *AdminClient) ServerUpdate(ctx context.Context, opts ServerUpdateOpts)
 
 	return us, nil
 }
+
+type NodeBumpVersionResp struct {
```
harshavardhana, line 92: Missing comments on these structs

vadmeste: done

vadmeste:
@klauspost can you please take a look ?

marktheunissen (approved):

info-commands.go
```diff
@@ -444,6 +446,8 @@ type ServerProperties struct {
 	License             *LicenseInfo      `json:"license,omitempty"`
 	IsLeader            bool              `json:"is_leader"`
 	ILMExpiryInProgress bool              `json:"ilm_expiry_in_progress"`
+	BackendVersion      uint32            `json:"backend_version"`
```
klauspost: Here you imply a number, but below the backend version is a semantic version. Seems like that should be aligned.

vadmeste: done

harshavardhana (approved):

PR #446, opened 2025-09-23
Adding storage capacity fields and node state indicators to cluster query API

zveinn:
EOS PR: https://github.com/miniohq/eos/pull/1617


klauspost:
@zveinn I don't understand your [DRAFT] and [IN PROGRESS] "tags"... But I don't CR them as long as they have these tags. In case that is unexpected, just drop the tags - at least for me it creates more confusion than it helps.

harshavardhana (approved):

PR #450, opened 2025-09-25
Adding a new "ClusterSummary" api + adding small summaries to Node and Drive queries.

zveinn:
related PR: https://github.com/miniohq/eos/pull/1643

### DIFF
--- a/info-v4-resources_gen.go
+++ b/info-v4-resources_gen.go
@@ -4120,12 +4120,6 @@ func (z *PaginatedDrivesResponse) DecodeMsg(dc *msgp.Reader) (err error) {
 				}
 			}
 			zb0001Mask |= 0x1
-		case "sum":
-			err = z.Summary.DecodeMsg(dc)
-			if err != nil {
-				err = msgp.WrapError(err, "Summary")
-				return
-			}
 		case "c":
 			z.Count, err = dc.ReadInt()
 			if err != nil {
@@ -4186,8 +4180,8 @@ func (z *PaginatedDrivesResponse) DecodeMsg(dc *msgp.Reader) (err error) {
 // EncodeMsg implements msgp.Encodable
 func (z *PaginatedDrivesResponse) EncodeMsg(en *msgp.Writer) (err error) {
 	// check for omitted fields
-	zb0001Len := uint32(8)
-	var zb0001Mask uint8 /* 8 bits */
+	zb0001Len := uint32(7)
+	var zb0001Mask uint8 /* 7 bits */
 	_ = zb0001Mask
 	if z.Results == nil {
 		zb0001Len--
@@ -4219,16 +4213,6 @@ func (z *PaginatedDrivesResponse) EncodeMsg(en *msgp.Writer) (err error) {
 					return
 				}
 			}
-		}
-		// write "sum"
-		err = en.Append(0xa3, 0x73, 0x75, 0x6d)
-		if err != nil {
-			return
-		}
-		err = z.Summary.EncodeMsg(en)
-		if err != nil {
-			err = msgp.WrapError(err, "Summary")
-			return
 		}
 		// write "c"
 		err = en.Append(0xa1, 0x63)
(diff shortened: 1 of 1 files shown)


info-v4-resources.go
```diff
@@ -210,6 +230,94 @@ type ErasureSetResource struct {
...
+	TotalVersions      int `json:"totalVersions" msg:"tv"`
+	ErasureSets        int `json:"erasureSets" msg:"es"`
+	DrivesPerSet       int `json:"drivesPerSet" msg:"dps"`
+	Parity             int `json:"parity" msg:"p"`
+}
+
+// PoolSummary contains summary information for a storage pool including usage and drive statistics
+type PoolSummary struct {
+	Index   int                 `json:"index" msg:"idx"`
+	Usage   ClusterSummaryUsage `json:"usage" msg:"us"`
```
cesnietor, line 269: would it make sense to call ClusterSummaryUsage -> `SummaryUsage` so that is more general? cause the current one inside PoolSummary makes it feel as if the pool is returning cluster info but it is not.

info-v4-resources.go
```diff
@@ -210,6 +230,94 @@ type ErasureSetResource struct {
...
+type ClusterSummaryResponse struct {
+	Encryption   bool                `json:"encryption" msg:"enc"`
+	Version      string              `json:"version" msg:"ver"`
+	DeploymentID string              `json:"deploymentID" msg:"did"`
+	Region       string              `json:"region" msg:"reg"`
+	Domains      []string            `json:"domains" msg:"dom"`
+	Mode         string              `json:"mode" msg:"mod"`
+	Usage        ClusterSummaryUsage `json:"usage" msg:"us"`
+	Servers      ClusterSummaryCount `json:"servers" msg:"srv"`
+	Drives       ClusterSummaryCount `json:"drives" msg:"drv"`
```
cesnietor: should `Drives` be `DriveSummaryCount` ? and then maybe we can have the `ClusterSummaryCount` no longer need Healing field for servers?, if it was to separate which struct belongs to which query then it's fine.

info-v4-resources.go
```diff
@@ -47,29 +47,52 @@ type PaginatedPoolsResponse struct {
...
+	Sort         string             `json:"sort" msg:"s"`
+	SortReversed bool               `json:"sortReversed" msg:"sr"`
 
 	// Aggregated are the metrics aggregated for all filtered drives,
 	// not just the results.
 	// Always returned, though day metrics are only available if the option is set.
 	Aggregated DiskMetric `json:"aggregated,omitempty" msg:"m,omitempty"`
 }
 
+// DrivesQuerySummary contains a summary for all drives, ignoring pagination and query limits.
```
cesnietor, line 82: just a thought here but since DrivesQuerySummary and Aggregated ignore pagination I was wondering if we should create another returning struct that contains both the PaginatedDrivesResponse and this new ones. Just to encapsulate how they work and make it self explanatory.
```
type DrivesQueryResponse struct {
    PaginatedResults PaginatedDrivesResponse
    Summary DrivesQuerySummary
    Aggregated DiskMetric `json:"aggregated,omitempty" msg:"m,omitempty"`
}
```

similar for Nodes.

cesnietor (commented):
Thanks Sveinn!, just left minor comments I leave those to you.

info-v4-resources.go
```diff
@@ -47,29 +47,52 @@ type PaginatedPoolsResponse struct {
...
 // PaginatedDrivesResponse represents a paginated response for drives
 type PaginatedDrivesResponse struct {
-	Results      []DriveResource `json:"results" msg:"r,omitempty"`
-	Count        int             `json:"count" msg:"c"`
-	Total        int             `json:"total" msg:"t"`
-	Offset       int             `json:"offset" msg:"o"`
-	Sort         string          `json:"sort" msg:"s"`
-	SortReversed bool            `json:"sortReversed" msg:"sr"`
+	Results      []DriveResource    `json:"results" msg:"r,omitempty"`
+	Summary      DrivesQuerySummary `json:"summary" msg:"sum"`
```
klauspost: This is already in `Aggregated`.

info-v4-resources.go
```diff
@@ -210,6 +230,94 @@ type ErasureSetResource struct {
...
+	Healing int `json:"healing" msg:"hl"`
+}
+
+// DriveSummaryCount contains drive counts with status breakdown
+type DriveSummaryCount struct {
+	Total   int `json:"total" msg:"t"`
+	Online  int `json:"online" msg:"on"`
+	Offline int `json:"offline" msg:"off"`
+	Healing int `json:"healing" msg:"hl"`
+}
```
klauspost, line 253: These would be covered by metrics

info-v4-resources.go
zveinn: poteito/potato

info-v4-resources.go
zveinn: I will be removing the DrivesQuerySummary since it's already contained inside aggregated.

PR #452, opened 2025-09-26
Add optional metrics to resource queries

klauspost:
Adds metrics to pool/nodes queries. Metrics will be aggregated for total result set as well as per entry.

Adds option to aggregate disk per set, so caller can reduce impact of queries.


harshavardhana (approved):

harshavardhana:
PTAL @cesnietor

harshavardhana:
PTAL @cesnietor

PR #462, opened 2025-10-07
making sort non-case sensitive

zveinn:


info-v4-resources.go
```diff
@@ -683,9 +683,26 @@ func SortSlice[T any](slice []T, field string, reversed bool) {
 		return
 	}
 
+	// findFieldByNameCaseInsensitive performs a case-insensitive field lookup
+	findFieldByNameCaseInsensitive := func(v reflect.Value, name string) reflect.Value {
```
harshavardhana: Really we need to do this ? Why can't we have a typed function here ?

zveinn: ask claude :S .. no I'm joking.. I didn't want to spend to much time on this. I'll clean it up right now.

zveinn: turns out, claude sonnet is much cleaner then opus. Had it refactor the function and it's much better

info-v4-resources.go
```diff
@@ -713,119 +713,130 @@ func SortSlice[T any](slice []T, field string, reversed bool) {
...
-		for v.Kind() == reflect.Ptr {
-			if v.IsNil() {
-				return reflect.Value{}
-			}
-			v = v.Elem()
+	parts := strings.Split(field, ".")
+	sort.SliceStable(slice, func(i, j int) bool {
+		// valI := unwrapPointer(reflect.ValueOf(slice[i]))
+		// valJ := unwrapPointer(reflect.ValueOf(slice[j]))
+
```
```suggestion
```

harshavardhana (approved):

PR #473, opened 2025-10-17
Returns the maximum Incoming and OutgoingBytes across all time segments.

SSushmitha8:
Required for  https://github.com/miniohq/aistor-console/issues/4782 Max column in the Nodewise metrics

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -1347,16 +1347,26 @@ func (a APIMetrics) LastMinuteTotal() APIStats {
 
 // LastDayTotalSegmented returns the total SegmentedAPIMetrics for the last day.
 // There will be no node-count for values.
-func (a APIMetrics) LastDayTotalSegmented() SegmentedAPIMetrics {
+// Also returns the maximum IncomingBytes and OutgoingBytes across all time segments.
+func (a APIMetrics) LastDayTotalSegmented() (SegmentedAPIMetrics, int64, int64) {
 	var res SegmentedAPIMetrics
 	for _, stats := range a.LastDayAPI {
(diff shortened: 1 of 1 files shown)


metrics.go
```diff
@@ -1360,16 +1347,26 @@ func (a APIMetrics) LastMinuteTotal() APIStats {
 
 // LastDayTotalSegmented returns the total SegmentedAPIMetrics for the last day.
 // There will be no node-count for values.
-func (a APIMetrics) LastDayTotalSegmented() SegmentedAPIMetrics {
+// Also returns the maximum IncomingBytes and OutgoingBytes across all time segments.
+func (a APIMetrics) LastDayTotalSegmented() (SegmentedAPIMetrics, int64, int64) {
```
harshavardhana, line 1351: Any reason why we can't have these two values as members of  res before the return? And have this set from the server side directly?

klauspost: Generally we want `SegmentedAPIMetrics` to remain lean and not have duplicate data as this is both in memory and on the wire.

That said the values are very arbitrary. This will allow you to get it on the returned value:

```go
// TrafficMax returns the maximum incoming and outgoing traffic across all segments.
func (s SegmentedAPIMetrics) TrafficMax() (maxIn, maxOut int64) {
	for _, segment := range s.Segments {
		maxIn = max(maxIn, segment.IncomingBytes)
		maxOut = max(maxOut, segment.OutgoingBytes)
	}
	return maxIn, maxOut
}
```

harshavardhana: yeah this is nicer and cleaner.

PR #477, opened 2025-10-20
Iam check

0xMALVEE:
issue: https://github.com/miniohq/eos/issues/1686
required for pr: https://github.com/miniohq/eos/pull/1947
base coppied from: https://github.com/minio/madmin-go/pull/464

### DIFF
--- a/user-commands.go
+++ b/user-commands.go
@@ -1000,7 +1000,13 @@ type IAMCacheAnalysis struct {
 
 // HasReplicationEntities returns true if there are any non-root IAM entities
 func (i IAMCacheAnalysis) HasReplicationEntities() bool {
-	return i != IAMCacheAnalysis{}
+	return i.TotalPolicies > 0 ||
+		i.TotalRegularUsers > 0 ||
+		i.TotalSvcAccNonRootParent > 0 ||
+		i.TotalGroups > 0 ||
+		i.TotalUserPolicyMappings > 0 ||
+		i.TotalGroupPolicyMappings > 0 ||
+		i.TotalSTSPolicyMappings > 0
 }
 
 // IAMEntityReport returns statistics about IAM cache entities


user-commands.go
```diff
@@ -985,3 +985,55 @@ func (adm *AdminClient) InfoAccessKey(ctx context.Context, accessKey string) (In
...
+// HasReplicationEntities returns true if there are any non-root IAM entities
+func (i IAMCacheAnalysis) HasReplicationEntities() bool {
+	return i.TotalPolicies > 0 ||
+		i.TotalRegularUsers > 0 ||
+		i.TotalSvcAccNonRootParent > 0 ||
+		i.TotalGroups > 0 ||
+		i.TotalUserPolicyMappings > 0 ||
+		i.TotalGroupPolicyMappings > 0 ||
+		i.TotalSTSPolicyMappings > 0
+}
```
klauspost, line 1010:
```suggestion
func (i IAMCacheAnalysis) HasReplicationEntities() bool {
	return i == IAMCacheAnalysis{}
```

user-commands.go
```diff
@@ -985,3 +985,55 @@ func (adm *AdminClient) InfoAccessKey(ctx context.Context, accessKey string) (In
...
+	data, err := DecryptData(adm.getSecretKey(), resp.Body)
+	if err != nil {
+		return IAMCacheAnalysis{}, err
+	}
+
+	var report IAMCacheAnalysis
+	if err = json.Unmarshal(data, &report); err != nil {
+		return IAMCacheAnalysis{}, err
+	}
+	return report, nil
```
klauspost:
```suggestion
	return report, json.Unmarshal(data, &report)
```

user-commands.go
taran-p: These are not equivalent since the previous code can return true even if TotalServiceAccounts > 0 (meaning if only root has service accounts) since those are not replicated.

PR #485, opened 2025-11-05
update: Add support of ongoing update cancellation

vadmeste:
Make it possible for the user to stop an ongoing cluster update. 
This will also stop rolling restart after the update if it is already started.


klauspost (approved):
I am a bit more questioning of how this will be on the server-side. Seems like something that will be completely untestable since the cluster can be in a billion different states.

But oh well.

PR #487, opened 2025-11-06
Tweak replication metrics

klauspost:
* 400/500 errors are not really good coverage. Instead we just log each type success.

* Move latency to per target stats.

* Add separate proxy errors (or rather "OK") per type.

* Upgrade msgp


klauspost:
(temp draft, checking if it needs further tweaks)

PR #488, opened 2025-11-06
Add TraceTrackDelete support for purge-on-delete tracing

poornas:
- Add TraceTrackDelete trace type to TraceType enum

### DIFF
--- a/service-commands.go
+++ b/service-commands.go
@@ -170,6 +170,7 @@ type ServiceTraceOpts struct {
 	ILM               bool
 	KMS               bool
 	Formatting        bool
+	PurgeOnDelete     bool
 
 	OnlyErrors    bool
 	Threshold     time.Duration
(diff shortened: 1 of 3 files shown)


service-commands.go
```diff
@@ -170,6 +170,7 @@ type ServiceTraceOpts struct {
 	ILM               bool
 	KMS               bool
 	Formatting        bool
+	TrackDelete       bool
```
harshavardhana: Just call it what it is `PurgeOnDelete` Track is not the word I would use here..

shtripat: +1. The suggested name `PurgeOnDelete` makes more sense to me as well.

klauspost: Why isn't there just a "Replication" type? Having these split into atoms just seems counterproductive.

harshavardhana: that also makes sense but TrackDelete is something that conflicts..

poornas: changed to PurgeOnDelete

harshavardhana (approved):

PR #506, opened 2025-12-14
Add distributed cluster health check support

allanrogerr:
Needed for https://github.com/miniohq/eos/pull/2508

## Summary
- Add `Distributed` option to `HealthOpts` to enable distributed cluster health checks
- When enabled, verifies health from ALL nodes' perspectives via Grid RPC
- Adds `distributed=true` query parameter to the cluster health endpoint

## Test plan
- [ ] Verify existing health check functionality remains unchanged when `Distributed` is false
- [ ] Test distributed health check with multi-node cluster

### DIFF
--- a/cluster-health.go
+++ b/cluster-health.go
@@ -58,15 +58,15 @@ func (an *AnonymousClient) Healthy(ctx context.Context, opts HealthOpts) (result
 	if opts.ClusterRead {
 		return an.clusterReadCheck(ctx)
 	}
-	return an.clusterCheck(ctx, opts.Maintenance, opts.Distributed)
+	return an.clusterCheck(ctx, opts)
 }
 
-func (an *AnonymousClient) clusterCheck(ctx context.Context, maintenance, distributed bool) (result HealthResult, err error) {
+func (an *AnonymousClient) clusterCheck(ctx context.Context, opts HealthOpts) (result HealthResult, err error) {
 	urlValues := make(url.Values)
-	if maintenance {
+	if opts.Maintenance {
 		urlValues.Set(maintanenceURLParameterKey, "true")
 	}
-	if distributed {
+	if opts.Distributed {
 		urlValues.Set(distributedURLParameterKey, "true")
 	}
 


cluster-health.go
```diff
@@ -49,21 +50,25 @@ type HealthResult struct {
...
+	Distributed bool // Verify health from ALL nodes' perspectives via Grid RPC
 }
 
 // Healthy will hit `/minio/health/cluster` and `/minio/health/cluster/ready` anonymous APIs to check the cluster health
 func (an *AnonymousClient) Healthy(ctx context.Context, opts HealthOpts) (result HealthResult, err error) {
 	if opts.ClusterRead {
 		return an.clusterReadCheck(ctx)
 	}
-	return an.clusterCheck(ctx, opts.Maintenance)
+	return an.clusterCheck(ctx, opts.Maintenance, opts.Distributed)
```
klauspost: Maybe just send `opts`?

cluster-health.go
```diff
@@ -49,21 +50,25 @@ type HealthResult struct {
...
 
-func (an *AnonymousClient) clusterCheck(ctx context.Context, maintenance bool) (result HealthResult, err error) {
+func (an *AnonymousClient) clusterCheck(ctx context.Context, maintenance, distributed bool) (result HealthResult, err error) {
 	urlValues := make(url.Values)
 	if maintenance {
 		urlValues.Set(maintanenceURLParameterKey, "true")
 	}
+	if distributed {
+		urlValues.Set(distributedURLParameterKey, "true")
+	}
```
klauspost: Where is this picked up? I see that `distributedURLParameterKey` is a new key, but I don't see where it is being read?

cluster-health.go
allanrogerr: Done

cluster-health.go
allanrogerr: This is for https://github.com/miniohq/eos/pull/2508#issuecomment-3649087740 @klauspost

PR #511, opened 2025-12-18
Fix error handling in ServerHealthInfo

anjalshireesh:
Read the error response before closing the HTTP response body. Previously, closeResponse(resp) was called before
httpRespToErrorResponse(resp), causing the error message from the server to be lost since the body was already closed.


shtripat (approved):
lgtm

0xMALVEE (changes requested):
`httpRespToErrorResponse` already have `closeResponse`
we can remove the extra `closeResponse` bellow

shtripat (approved):

0xMALVEE (approved):

PR #515, opened 2025-12-30
fix: metrics error parsing

shtripat:
remove redundent closeResponse (causing draining of data before reading) since it's already present in httpRespToErrorResponse


anjalshireesh (approved):
LGTM

PR #519, opened 2026-01-12
Add 24h Runtime, Mem, Process, Scanner, OS and Net metrics

klauspost:
Adds fairly minimal persisted stats to most remaining types

Added to mnav, so will be in tui and `mc admin cluster stats` with dependency update.

### DIFF
--- a/mnav/mem.go
+++ b/mnav/mem.go
@@ -60,13 +60,14 @@ func (node *MemMetricsNavigator) GetChildren() []MetricChild {
 	if node.mem == nil {
 		return []MetricChild{}
 	}
-	children := []MetricChild{
-		{Name: "usage", Description: "Core memory usage statistics and utilization"},
-		{Name: "system", Description: "System memory details (cache, buffers, shared)"},
-		{Name: "swap", Description: "Swap space information and utilization"},
-		{Name: "limits", Description: "Memory limits and cgroup configuration"},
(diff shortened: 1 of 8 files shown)


metrics.go
```diff
@@ -1100,6 +1135,10 @@ func (n *NetMetrics) Merge(other *NetMetrics) {
 		}
 		n.Interfaces[k] = n.Interfaces[k].add(v)
 	}
+	if other.LastDay != nil && n.LastDay == nil {
```
vadmeste, line 1138: I think you mean:
```
if other.LastDay != nil {
    n.LastDay = new(SegmentedInterfaceStats)
    n.LastDay.Add(other.LastDay)
}

klauspost: No. We are merging 'other' into 'n'. That would overwrite it.

We just want to make sure that if 'other.LastDay' is set we have something to merge into.

vadmeste: yes correct, the suggestion is wrong, however since we are in Merge(), isn't safer to make sure we initialize n.LastDay if it is nil ?

currently n.LastDay.Add() in line 1141 is called any time

klauspost: No. We don't want it initialized if it doesn't have a value. That will just add needless allocs. The Add function will just ignore it if "other" is nil.

PR #521, opened 2026-01-13
inspect: Add pool parameter to extract data from a specific pool

vadmeste:
Useful to inspect leftover data from a decommissioned pool.

### DIFF
--- a/inspect.go
+++ b/inspect.go
@@ -27,12 +27,14 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 )
 
 // InspectOptions provides options to Inspect.
 type InspectOptions struct {
 	Volume, File string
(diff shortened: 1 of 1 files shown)


inspect.go
```diff
@@ -27,12 +27,14 @@ import (
...
 	"net/http"
 	"net/url"
+	"strconv"
 )
 
 // InspectOptions provides options to Inspect.
 type InspectOptions struct {
 	Volume, File string
 	PublicKey    []byte // PublicKey to use for inspected data.
+	PoolNumber   int    // Pool number (1-based) to extract data from. Use 0 for all pools.
```
klauspost: I was really hoping we could get rid of all the "1-based" in our API at least.

```suggestion
	PoolIdx   *int    // Pool to fetch. nil = fetch from all.
```

PR #522, opened 2026-01-13
Fix new linter issues

vadmeste:


klauspost:
Added in https://github.com/minio/madmin-go/pull/519

PR #528, opened 2026-01-29
feat: add alert types for config mismatch, erasure set health, KMS, and storage capacity

Praveenrajmani:
Introduces a new GetAlerts admin API (POST /admin/alerts) that streams alert events from the MinIO server as newline-delimited JSON, using Go 1.23's iter.Seq2 for idiomatic range-over-function iteration.

**New types:**

- Alert — the top-level alert event with type, timestamp, title, message, dedup key, and deployment/cluster metadata
- AlertDetails — strongly-typed union struct; only one field populated per alert type
- AlertLogOpts — query options (filter by type, check interval, max per node)

**Alert types and their detail structs:**

- license-expiry (LicenseExpiryDetails) — license ID, expiry time, state
- certificate-expiry (CertificateExpiryDetails) — common name, serial, DNS names, days until expiry
- config-mismatch (ConfigMismatchDetails) — peer address, mismatch type (env/args/endpoints), differing values
- erasure-set-health (ErasureSetHealthDetails) — pool/set ID, online/total drives, read+write quorum, condition
- kms-unavailable (KMSUnavailableDetails) — KMS type
- storage-capacity (StorageCapacityDetails) — free/total bytes, free percent

Also adds AlertInternalSubSys, AlertWebhookSubSys, and AlertKafkaSubSys subsystem constants to parse-config.go for alert routing configuration.

### DIFF
--- a/alert.go
+++ b/alert.go
@@ -1,0 +1,111 @@
+// Copyright (c) 2015-2026 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package madmin
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+	"iter"
+	"net/http"
+	"time"
+)
+
+// AlertType represents the type of alert event
+type AlertType string
+
+const (
+	// AlertTypeLicense represents license expiration alerts
+	AlertTypeLicense AlertType = "license-expiry"
+	// AlertTypeCertificate represents TLS certificate expiration alerts
+	AlertTypeCertificate AlertType = "certificate-expiry"
+	// AlertTypeConfigMismatch represents server configuration mismatch alerts
+	AlertTypeConfigMismatch AlertType = "config-mismatch"
+	// AlertTypeErasureSetHealth represents erasure set write-redundancy degradation alerts
+	AlertTypeErasureSetHealth AlertType = "erasure-set-health"
+	// AlertTypeKMSUnavailable represents KMS connectivity failure alerts
+	AlertTypeKMSUnavailable AlertType = "kms-unavailable"
+	// AlertTypeStorageCapacity represents storage capacity critical alerts
+	AlertTypeStorageCapacity AlertType = "storage-capacity"
+)
+
+// Alert represents a single alert event in the system.
+// It captures alert information with contextual metadata including deployment and cluster information.
+type Alert struct {
+	Type         AlertType         `json:"type"`
+	Timestamp    time.Time         `json:"timestamp"`
+	Title        string            `json:"title"`
+	Message      string            `json:"message"`
+	Details      map[string]string `json:"details,omitempty"`
+	DeploymentID string            `json:"deploymentId"`
+	ClusterName  string            `json:"clusterName"`
+	DedupKey     string            `json:"dedupKey,omitempty"`
+}
+
+// AlertLogOpts represents options for querying alerts
+type AlertLogOpts struct {
+	Types      []string      `json:"types,omitempty"`
+	Interval   time.Duration `json:"interval,omitempty"`
+	MaxPerNode int           `json:"maxPerNode,omitempty"`
+}
+
+// GetAlerts returns alerts stored in the system as a streaming JSON response
+// via POST /admin/alerts. Use AlertLogOpts.Interval to control the server-side
+// check interval.
+func (adm AdminClient) GetAlerts(ctx context.Context, opts AlertLogOpts) iter.Seq2[*Alert, error] {
+	return func(yield func(*Alert, error) bool) {
+		alertOpts, err := json.Marshal(opts)
+		if err != nil {
+			yield(nil, err)
+			return
+		}
+		reqData := requestData{
+			relPath: adminAPIPrefix + "/alerts",
+			content: alertOpts,
+		}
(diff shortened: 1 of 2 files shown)


alert.go
```diff
@@ -0,0 +1,126 @@
...
+type AlertLogOpts struct {
+	Type       string        `json:"type,omitempty"`
+	Interval   time.Duration `json:"interval,omitempty"`
+	MaxPerNode int           `json:"maxPerNode,omitempty"`
+}
+
+// GetAlerts retrieves alerts from the MinIO server with the specified options.
+// It returns an iterator that yields AlertEvent objects.
+// Type filtering is case-insensitive and handled server-side.
+func (adm AdminClient) GetAlerts(ctx context.Context, opts AlertLogOpts) iter.Seq2[AlertEvent, error] {
```
klauspost:
```suggestion
func (adm AdminClient) GetAlerts(ctx context.Context, opts AlertLogOpts) iter.Seq2[*AlertEvent, error] {
```
It will either return an error or a value. Cleaner to deal with, less copying.

parse-config.go
```diff
@@ -88,6 +88,10 @@ const (
 	LogAPIKafkaSubSys   = "log_api_kafka"
 	LogErrorKafkaSubSys = "log_error_kafka"
 	LogAuditKafkaSubSys = "log_audit_kafka"
+
+	AlertInternalSubSys = "alert_internal"
+	AlertWebhookSubSys  = "alert_webhook"
+	AlertKafkaSubSys    = "alert_kafka"
```
klauspost, line 99: Shouldn't these be included in AlertEvent?

I kinda assume "webhook" can return more than one alert.. or at least subsystems will be able to in the future.

alert.go
```diff
@@ -0,0 +1,126 @@
...
+	// AlertTypeLicense represents license expiration alerts
+	AlertTypeLicense AlertType = "license-expiry"
+	// AlertTypeCertificate represents TLS certificate expiration alerts
+	AlertTypeCertificate AlertType = "certificate-expiry"
+)
+
+// AlertEvent represents a single alert event in the system.
+// It captures alert information with contextual metadata including deployment and site information.
+type AlertEvent struct {
+	ID           string       `json:"id"`
```
klauspost: What is this ID? Should it deduplicate on this?

alert.go
```diff
@@ -0,0 +1,126 @@
...
+// AlertLogOpts represents options for querying alerts
+type AlertLogOpts struct {
+	Type       string        `json:"type,omitempty"`
+	Interval   time.Duration `json:"interval,omitempty"`
+	MaxPerNode int           `json:"maxPerNode,omitempty"`
+}
+
+// GetAlerts retrieves alerts from the MinIO server with the specified options.
+// It returns an iterator that yields AlertEvent objects.
+// Type filtering is case-insensitive and handled server-side.
```
klauspost: Will "current" events be returned right away? Will events re-fire at regular intervals? What be expected from the API?

alert.go
```diff
@@ -0,0 +1,126 @@
...
+	CommonName      string    `json:"commonName"`
+	SerialNumber    string    `json:"serialNumber"`
+	NotAfter        time.Time `json:"notAfter"`
+	DaysUntilExpiry int       `json:"daysUntilExpiry"`
+	DNSNames        []string  `json:"dnsNames"`
+}
+
+// AlertLogOpts represents options for querying alerts
+type AlertLogOpts struct {
+	Type       string        `json:"type,omitempty"`
```
klauspost: You can only monitor one at the time?

parse-config.go
Praveenrajmani: The AlertInternalSubSys, AlertWebhookSubSys, AlertKafkaSubSys are configuration subsystems

This is required for a feature in EOS to send alerts to configured targets / internally;

alert.go
Praveenrajmani: ID can be removed

alert.go
Praveenrajmani: GetAlerts is a streaming API that:

- Makes a POST request to /admin/alerts
- Returns alerts as a stream decoded from JSON
- Uses Interval option to control the check interval on server side

alert.go
Praveenrajmani: changed this to []string

alert.go
klauspost: I was implying it should be documented.

parse-config.go
klauspost: I was thinking in the `type Alert struct`.

Praveenrajmani: yeah that is the payload struct and these are sub configs which are required for the KVS; unifying it will be not efficient as these are multi-targeted subsystems

klauspost: I don't understand. But whatever,

alert.go
```diff
@@ -0,0 +1,171 @@
...
+	KMSUnavailable    *KMSUnavailableDetails    `json:"kmsUnavailable,omitempty"`
+	StorageCapacity   *StorageCapacityDetails   `json:"storageCapacity,omitempty"`
+}
+
+// ConfigMismatchDetails contains server configuration mismatch alert information.
+type ConfigMismatchDetails struct {
+	PeerAddress    string   `json:"peerAddress"`
+	MismatchType   string   `json:"mismatchType"` // "env", "args", "endpoints"
+	ExpectedCount  int      `json:"expectedCount,omitempty"`
+	ActualCount    int      `json:"actualCount,omitempty"`
```
shtripat: are these two expected and actual ENV counts? if so, can we rename them like that?

alert.go
```diff
@@ -0,0 +1,171 @@
...
+	ExpiresAt time.Time `json:"expiresAt"`
+	State     string    `json:"state"` // "expiring_soon", "grace_period", "read_only", "fully_expired"
+}
+
+// CertificateExpiryDetails contains certificate-specific alert information.
+type CertificateExpiryDetails struct {
+	CommonName      string    `json:"commonName"`
+	SerialNumber    string    `json:"serialNumber"`
+	NotAfter        time.Time `json:"notAfter"`
+	DaysUntilExpiry int       `json:"daysUntilExpiry"`
```
shtripat: instead cant we just have `ExpiresAt`?

alert.go
Praveenrajmani: these are endpoints count, have changed it

alert.go
Praveenrajmani: these fields will be displayed to customer, daysUntilExpiry is more customer friendly to know how many days

shtripat (approved):
looks good overall

dhananjaykrutika (approved):

PR #530, opened 2026-01-30
Add parquet tags to API log struct

Praveenrajmani:
Add parquet struct tags to API and CallInfo types to enable Parquet file serialization for API logs.

### DIFF
--- a/log/api.go
+++ b/log/api.go
@@ -53,43 +53,43 @@ const (
 
 // API represents the api event
 type API struct {
-	Version   string            `json:"version"`
-	Time      time.Time         `json:"time"`
-	Node      string            `json:"node,omitempty"`
-	Origin    Origin            `json:"origin,omitempty"`
-	Type      APIType           `json:"type,omitempty"`
-	Name      string            `json:"name,omitempty"`
-	Bucket    string            `json:"bucket,omitempty"`
-	Object    string            `json:"object,omitempty"`
-	VersionID string            `json:"versionId,omitempty"`
-	Tags      map[string]string `json:"tags,omitempty"`
-	CallInfo  *CallInfo         `json:"callInfo,omitempty"`
+	Version   string            `json:"version" parquet:"version"`
+	Time      time.Time         `json:"time" parquet:"time,timestamp(microsecond)"`
+	Node      string            `json:"node,omitempty" parquet:"node,optional"`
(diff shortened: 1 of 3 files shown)


log/api.go
```diff
@@ -53,43 +53,43 @@ const (
...
+	RequestTime       string            `json:"requestTime,omitempty" parquet:"requestTime,optional"`
+	TimeToResponse    string            `json:"timeToResponse,omitempty" parquet:"timeToResponse,optional"`
+	ReadBlocked       string            `json:"readBlocked,omitempty" parquet:"readBlocked,optional"`
+	WriteBlocked      string            `json:"writeBlocked,omitempty" parquet:"writeBlocked,optional"`
+	SourceHost        string            `json:"sourceHost,omitempty" parquet:"sourceHost,optional"`
+	RequestID         string            `json:"requestID,omitempty" parquet:"requestId,optional"`
+	UserAgent         string            `json:"userAgent,omitempty" parquet:"userAgent,optional"`
+	ReqPath           string            `json:"requestPath,omitempty" parquet:"reqPath,optional"`
+	ReqHost           string            `json:"requestHost,omitempty" parquet:"reqHost,optional"`
+	ReqClaims         map[string]string `json:"requestClaims,omitempty" parquet:"reqClaims,optional"`
```
klauspost, line 87: This smells like a breaking change. Do you guarantee that older versions aren't going to send a number as value?

Praveenrajmani: the API log feature is not released yet @klauspost 

And we are safe here, this is how I construct this in EOS (reqClaims is of `map[string]any` by native, I wanted to convert it to map[string]string to make it parquet-friendly)

```go
func claimsToStringMap(claims map[string]any) map[string]string {
	if claims == nil {
		return nil
	}
	result := make(map[string]string, len(claims))
	for k, v := range claims {
		if v == nil {
			continue
		}
		switch val := v.(type) {
		case string:
			result[k] = val
		default:
			result[k] = fmt.Sprintf("%v", val)
		}
	}
	return result
}
```

klauspost: @Praveenrajmani Isn't it released now? I can't see how it can be safe if current version can send anything and the receiver only accepts string?

Praveenrajmani: Yeah, the feature using this struct is still on "edge". It is not released yet @klauspost

Praveenrajmani:
https://github.com/miniohq/eos/pull/3180 is the PR @klauspost

PR #533, opened 2026-02-06
Add table maintenance API req/rsp structs based on S3 Tables APIs

dhananjaykrutika:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_s3Buckets_PutTableMaintenanceConfiguration.html
https://docs.aws.amazon.com/AmazonS3/latest/API/API_s3Buckets_GetTableMaintenanceConfiguration.html

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,8 @@ linters:
         text: error strings should not be capitalized or end with punctuation or a newline
       - path: (.+)\.go$
         text: don't use ALL_CAPS in Go names
+      - path: (.+)\.go$
+        text: "var-naming: avoid package names that conflict with Go standard library package names"
     paths:
       - third_party$
       - builtin$
--- a/kms-commands.go
+++ b/kms-commands.go
@@ -129,6 +129,21 @@ type KMSMetrics struct {
 	StackAlloc  int64 `json:"kes_system_mem_stack_used"`
 }
 
+// EnableKMS enables the internal builtin KMS on a MinIO cluster.
+// It returns an error if the builtin KMS is already enabled.
+func (adm *AdminClient) EnableKMS(ctx context.Context) error {
+	// POST /minio/kms/v1/enable
+	resp, err := adm.doKMSRequest(ctx, "/enable", http.MethodPost, nil, map[string]string{})
+	if err != nil {
+		return err
+	}
+	defer closeResponse(resp)
+	if resp.StatusCode != http.StatusOK {
+		return httpRespToErrorResponse(resp)
+	}
+	return nil
+}
+
 type KMSAPI struct {
 	Method  string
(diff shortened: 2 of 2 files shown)


table-maintenance.go
```diff
@@ -0,0 +1,75 @@
...
+	MinSnapshotsToKeep *int `json:"minSnapshotsToKeep,omitempty"`
+}
+
+// TableMaintenanceSettings is a union type containing maintenance settings.
+// Only one of the fields should be set at a time based on the maintenance type.
+type TableMaintenanceSettings struct {
+	IcebergSnapshotManagement *IcebergSnapshotManagementSettings `json:"icebergSnapshotManagement,omitempty"`
+	// IcebergCompaction is a placeholder for future compaction settings.
+	// Currently not implemented but reserved for API compatibility.
+	IcebergCompaction *json.RawMessage `json:"icebergCompaction,omitempty"`
```
klauspost: Why is it we don't add the [existing type](https://docs.aws.amazon.com/AmazonS3/latest/API/API_s3Buckets_IcebergCompactionSettings.html)?

Without more details, it seems strictly worse to add it like this than omitting it, since you would inevitably break API compatibility *when* you add it.

(also you shouldn't need a pointer to `json.RawMessage` since it has a slice that can be nil).

table-maintenance.go
```diff
@@ -0,0 +1,75 @@
...
+// Table maintenance type constants
+const (
+	MaintenanceTypeIcebergSnapshotManagement = "icebergSnapshotManagement"
+	MaintenanceTypeIcebergCompaction         = "icebergCompaction"
+)
+
+// Table maintenance status constants
+const (
+	MaintenanceStatusEnabled  = "enabled"
+	MaintenanceStatusDisabled = "disabled"
```
klauspost: I suggest you add a `type MaintenanceStatus string` and make these typed.

table-maintenance.go
```diff
@@ -0,0 +1,75 @@
...
+//
+
+package madmin
+
+import "encoding/json"
+
+// Table maintenance type constants
+const (
+	MaintenanceTypeIcebergSnapshotManagement = "icebergSnapshotManagement"
+	MaintenanceTypeIcebergCompaction         = "icebergCompaction"
```
klauspost, line 27: I don't see where/how these are used?

table-maintenance.go
dhananjaykrutika: That makes sense. I will drop the field for compaction for now.

table-maintenance.go
dhananjaykrutika: These are used in `eos` as part of the actual implementation of the API and in `ec` as part of the command-line based maintenance job configuration option.

PR #534, opened 2026-02-11
add functions for enabling the builtin KMS and rotating keys

aead:
This commit adds two methods:
```
- EnableKMS  // Enables the builtin KMS on a MinIO cluster
- RotateKey  // Creates a new key version
```

It also adds a `KeyVersion` field to the `KMSKeyStatus` struct.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,8 @@ linters:
         text: error strings should not be capitalized or end with punctuation or a newline
       - path: (.+)\.go$
         text: don't use ALL_CAPS in Go names
+      - path: (.+)\.go$
+        text: "var-naming: avoid package names that conflict with Go standard library package names"
     paths:
       - third_party$
       - builtin$
--- a/kms-commands.go
+++ b/kms-commands.go
@@ -129,6 +129,21 @@ type KMSMetrics struct {
 	StackAlloc  int64 `json:"kes_system_mem_stack_used"`
 }
 
+// EnableKMS enables the internal builtin KMS on a MinIO cluster.
+// It returns an error if the builtin KMS is already enabled.
+func (adm *AdminClient) EnableKMS(ctx context.Context) error {
+	// POST /minio/kms/v1/enable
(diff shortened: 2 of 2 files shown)


.golangci.yml
```diff
@@ -21,6 +21,8 @@ linters:
         text: error strings should not be capitalized or end with punctuation or a newline
       - path: (.+)\.go$
         text: don't use ALL_CAPS in Go names
+      - path: (.+)\.go$
+        text: "var-naming: avoid package names that conflict with Go standard library package names"
```
klauspost, line 25: Accident? New check?
```suggestion
```

kms-commands.go
```diff
@@ -231,6 +246,25 @@ func (adm *AdminClient) CreateKey(ctx context.Context, keyID string) error {
 	return nil
 }
 
+// RotateKey creates a new version of the master key with the given key ID\
+// at a MinIO cluster.
+func (adm *AdminClient) RotateKey(ctx context.Context, keyID string) error {
```
klauspost, line 251: Couldn't/shouldn't we make this idempotent?

.golangci.yml
aead: CI failed since `madmin/log` has the same name as stdlib `log` - I added this on purpose. Removing this causes revive to complain

kms-commands.go
aead: How? Key rotation means adding another key (version) to an existing key ring.
For some master key `key-1` you have the something like this:
```
{
  "1": <key-X>
  "2": <key-Y>
}
```
After rotation you have:
```
{
  "1": <key-X>
  "2": <key-Y>
  "3": <key-Z>
}
```

klauspost: I guess it is not the end of the world to have a few extra keys.

.golangci.yml
klauspost: Ah. It is package naming. Yeah. We can't really fix that now.

PR #571, opened 2026-04-13
Add capacity forecast client method

cniackz:
Adds the CapacityForecast struct and the AdminClient.CapacityForecast method that calls GET /admin/v4/capacity-forecast and returns the projection. Days-until threshold fields are pointer floats where nil means no prediction is possible and a negative value means the threshold was already crossed in the past, so magnitude is meaningful. Follows the existing ServerInfo client pattern.

### DIFF
--- a/capacity-forecast.go
+++ b/capacity-forecast.go
@@ -35,6 +35,18 @@ type CapacityForecast struct {
 	DaysUntil100Pct    float64 `json:"daysUntil100Pct"`
 	GrowthRatePerDay   int64   `json:"growthRatePerDay"`
 	DataPointCount     int     `json:"dataPointCount"`
+
+	// Worst-case prediction based on the largest single-day growth
+	// observed between any two consecutive data points.
+	MinDaysUntilFull float64 `json:"minDaysUntilFull"`
+
+	// Confidence metrics for the linear regression.
+	RSquared float64 `json:"rSquared"` // 0-1, goodness of fit
+	Variance float64 `json:"variance"` // variance of daily usedFraction deltas
+
+	// Short-window (14-day) regression for recency-weighted predictions.
+	RecentGrowthRatePerDay float64 `json:"recentGrowthRatePerDay"`
+	RecentDaysUntilFull    float64 `json:"recentDaysUntilFull"`
 }
 
 // CapacityForecast returns a storage capacity forecast based on
--- a/capacity-forecast.go
+++ b/capacity-forecast.go
@@ -44,6 +44,11 @@ type CapacityForecast struct {
 	RSquared float64 `json:"rSquared"` // 0-1, goodness of fit
 	Variance float64 `json:"variance"` // variance of daily usedFraction deltas
 
+	// Concrete min/max daily changes in usedFraction between consecutive
+	// data points. DayMinDelta can be negative (space was freed).
+	DayMinDelta float64 `json:"dayMinDelta"`
+	DayMaxDelta float64 `json:"dayMaxDelta"`
+
 	// Short-window (14-day) regression for recency-weighted predictions.
 	RecentGrowthRatePerDay float64 `json:"recentGrowthRatePerDay"`
 	RecentDaysUntilFull    float64 `json:"recentDaysUntilFull"`
--- a/capacity-forecast.go
+++ b/capacity-forecast.go
@@ -26,19 +26,26 @@ import (
 
 // CapacityForecast contains storage capacity predictions based on
 // historical daily snapshots collected by the scanner.
+//
+// Days-until-threshold fields are pointers: nil means "unknown" (for example,
+// not enough history yet, or usage is not growing). A concrete value may be
+// negative when the threshold was already crossed in the past.
 type CapacityForecast struct {
 	CurrentUsedBytes   uint64  `json:"currentUsedBytes"`
 	CurrentTotalBytes  uint64  `json:"currentTotalBytes"`
 	CurrentUsedPercent float64 `json:"currentUsedPercent"`
-	DaysUntil80Pct     float64 `json:"daysUntil80Pct"`
-	DaysUntil90Pct     float64 `json:"daysUntil90Pct"`
-	DaysUntil100Pct    float64 `json:"daysUntil100Pct"`
-	GrowthRatePerDay   int64   `json:"growthRatePerDay"`
-	DataPointCount     int     `json:"dataPointCount"`
+
+	// Days until each usage threshold is reached. nil = unknown.
+	DaysUntil80Pct  *float64 `json:"daysUntil80Pct,omitempty"`
+	DaysUntil90Pct  *float64 `json:"daysUntil90Pct,omitempty"`
+	DaysUntil100Pct *float64 `json:"daysUntil100Pct,omitempty"`
+
+	GrowthRatePerDay int64 `json:"growthRatePerDay"`
+	DataPointCount   int   `json:"dataPointCount"`
 
 	// Worst-case prediction based on the largest single-day growth
-	// observed between any two consecutive data points.
-	MinDaysUntilFull float64 `json:"minDaysUntilFull"`
+	// observed between any two consecutive data points. nil = unknown.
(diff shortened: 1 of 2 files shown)


capacity-forecast.go
```diff
@@ -0,0 +1,63 @@
...
+
+// CapacityForecast contains storage capacity predictions based on
+// historical daily snapshots collected by the scanner.
+type CapacityForecast struct {
+	CurrentUsedBytes   uint64  `json:"currentUsedBytes"`
+	CurrentTotalBytes  uint64  `json:"currentTotalBytes"`
+	CurrentUsedPercent float64 `json:"currentUsedPercent"`
+	DaysUntil80Pct     int     `json:"daysUntil80Pct"`
+	DaysUntil90Pct     int     `json:"daysUntil90Pct"`
+	DaysUntil100Pct    int     `json:"daysUntil100Pct"`
```
klauspost: Assuming you get a float value from your calculation. Keep that and let the client determine how to show it.

capacity-forecast.go
```diff
@@ -0,0 +1,63 @@
...
+	GrowthRatePerDay   int64   `json:"growthRatePerDay"`
```
klauspost: Add minimum value of above 4 values, which uses the biggest delta of any two days and uses that to predict exhaustion.

capacity-forecast.go
```diff
@@ -0,0 +1,63 @@
...
+	DataPointCount     int     `json:"dataPointCount"`
```
klauspost: Add variance, say the average growth is 100GB/day. If minimum is -10GB/day and max is 200MB/day.

DayMin = -10%
DayMax = 100%

Probably we also want a more easily understandable "confidence factor" - whether it is stddev from the linear regression or something else. Some of these data will be VERY chaotic, so we need to express that properly.

capacity-forecast.go
cniackz: Added `MinDaysUntilFull` — finds the largest single-day increase in usedFraction between consecutive data points and uses that as the growth rate to predict exhaustion. Returns -1 if no positive delta exists.

capacity-forecast.go
cniackz: Implemented all three:

1. **DayMin/DayMax** — added `DayMinDelta` and `DayMaxDelta` (float64) representing the smallest and largest single-day change in usedFraction. Negative values indicate space was freed.

2. **Variance** — computed from day-to-day deltas via `computeDeltaStats()` (single-pass over consecutive points).

3. **Confidence** — `RSquared` (0-1 goodness of fit) exposed as raw value. EC displays it as high/medium/low with the R² value.

bazko1 (approved):

capacity-forecast.go
```diff
@@ -0,0 +1,87 @@
...
+	CurrentTotalBytes  uint64  `json:"currentTotalBytes"`
+	CurrentUsedPercent float64 `json:"currentUsedPercent"`
+
+	// Days until each usage threshold is reached. nil = unknown.
+	DaysUntil80Pct  *float64 `json:"daysUntil80Pct,omitempty"`
+	DaysUntil90Pct  *float64 `json:"daysUntil90Pct,omitempty"`
+	DaysUntil100Pct *float64 `json:"daysUntil100Pct,omitempty"`
+
+	GrowthRatePerDay int64 `json:"growthRatePerDay"`
+	DataPointCount   int   `json:"dataPointCount"`
```
klauspost: Is `DataPointCount` related to `GrowthRatePerDay`? So is it the difference between now and now-DataPointCount?

I feel these variables could be named better.

capacity-forecast.go
```diff
@@ -0,0 +1,87 @@
...
+// negative when the threshold was already crossed in the past.
+type CapacityForecast struct {
+	CurrentUsedBytes   uint64  `json:"currentUsedBytes"`
+	CurrentTotalBytes  uint64  `json:"currentTotalBytes"`
+	CurrentUsedPercent float64 `json:"currentUsedPercent"`
+
+	// Days until each usage threshold is reached. nil = unknown.
+	DaysUntil80Pct  *float64 `json:"daysUntil80Pct,omitempty"`
+	DaysUntil90Pct  *float64 `json:"daysUntil90Pct,omitempty"`
+	DaysUntil100Pct *float64 `json:"daysUntil100Pct,omitempty"`
```
klauspost: Is this long/short/mid term trend? Describe it in one or two sentences.

capacity-forecast.go
```diff
@@ -0,0 +1,87 @@
...
+	MinDaysUntilFull *float64 `json:"minDaysUntilFull,omitempty"`
+
+	// Confidence from Kalman filter covariance (0-1, higher = more confident).
+	RSquared float64 `json:"rSquared"`
+	Variance float64 `json:"variance"` // variance of daily usedFraction deltas
+
+	// Concrete min/max daily changes in usedFraction between consecutive
+	// data points. DayMinDelta can be negative (space was freed).
+	DayMinDelta float64 `json:"dayMinDelta"`
+	DayMaxDelta float64 `json:"dayMaxDelta"`
```
klauspost: Percentage or bytes?

klauspost (commented):
Let's see if we can qualify the numbers a bit more.

capacity-forecast.go
cniackz: Thank you @klauspost, you are right. They are not related. Put a better name to both of them.

capacity-forecast.go
cniackz: Thank you @klauspost. I have put a comment in these fields that explains better in couple of sentences what they are and how they work, please look at the commit: e09bc75

capacity-forecast.go
cniackz: It was expressed as a fraction in [-1, +1] but it got changed to bytes. For example, imagine we have these deltas in different days: (+2, +2, -1, +2, +5, 0 TB) then `DayMinDeltaBytes` is `-1099511627776` (1 TiB Freed) and `DayMaxDeltaBytes` is `+5497558138880` (5 TiB spike).

capacity-forecast.go
```diff
@@ -0,0 +1,99 @@
...
+// Days-until-threshold fields are pointers: nil means "unknown" (for example,
+// not enough history yet, or usage is not growing). A concrete value may be
+// negative when the threshold was already crossed in the past.
+type CapacityForecast struct {
+	CurrentUsedBytes   uint64  `json:"currentUsedBytes"`
+	CurrentTotalBytes  uint64  `json:"currentTotalBytes"`
+	CurrentUsedPercent float64 `json:"currentUsedPercent"`
+
+	// Days until each usage threshold is reached. The projection comes from
+	// a Kalman filter that integrates up to 365 daily samples with greater
```
klauspost, line 39: Wasn't it 90 days/samples? That should be more than plenty.

Kalman has a pretty high recency bias - I wouldn't expect it to produce any different result from 365.

cniackz: @klauspost the idea is to predict also for companies which flow is longer than 90 days. Need to talk to Kannappan if we need to revert this back to 90.

capacity-forecast.go
```diff
@@ -0,0 +1,99 @@
...
+	// Recency-weighted predictions from the Kalman filter.
+	RecentGrowthRatePerDay float64  `json:"recentGrowthRatePerDay"`
+	RecentDaysUntilFull    *float64 `json:"recentDaysUntilFull,omitempty"`
+}
+
+// CapacityForecast returns a storage capacity forecast based on
+// historical daily snapshots.
+func (adm *AdminClient) CapacityForecast(ctx context.Context) (CapacityForecast, error) {
+	resp, err := adm.executeMethod(ctx,
+		http.MethodGet,
```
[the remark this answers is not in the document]
cniackz: Added httptest coverage in commit a70ee5b: validates GET path, JSON decoding for nil/positive/negative `DaysUntil*` pointer fields, and the non-200 error path. 6 tests passing. Thank you 🙏

harshavardhana (approved):

PR #574, opened 2026-04-20
Add IcebergUnreferencedFileRemoval maintenance type and settings

dhananjaykrutika:
The spec is based on https://docs.aws.amazon.com/AmazonS3/latest/API/API_s3Buckets_IcebergUnreferencedFileRemovalSettings.html

### DIFF
--- a/table-maintenance.go
+++ b/table-maintenance.go
@@ -23,8 +23,9 @@ package madmin
 
 // Table maintenance type constants
 const (
-	MaintenanceTypeIcebergSnapshotManagement = "icebergSnapshotManagement"
-	MaintenanceTypeIcebergCompaction         = "icebergCompaction"
+	MaintenanceTypeIcebergSnapshotManagement      = "icebergSnapshotManagement"
+	MaintenanceTypeIcebergCompaction              = "icebergCompaction"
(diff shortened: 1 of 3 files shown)


table-maintenance.go
```diff
@@ -53,11 +54,20 @@ type IcebergCompactionSettings struct {
 	TargetFileSizeMB *int `json:"targetFileSizeMB,omitempty"`
 }
 
+// IcebergUnreferencedFileRemovalSettings contains settings for Iceberg unreferenced file removal.
+// Files are eligible to be marked noncurrent when their age (anchored to creation time) exceeds
+// UnreferencedDays. Once noncurrent, files are permanently deleted after NoncurrentDays.
+type IcebergUnreferencedFileRemovalSettings struct {
+	UnreferencedDays *int `json:"unreferencedDays,omitempty"`
+	NoncurrentDays   *int `json:"noncurrentDays,omitempty"`
```
klauspost, line 62: Why are these pointers? Does it fall back to some default or something?

dhananjaykrutika: Yes, that is correct.

marktheunissen (approved):

harshavardhana (approved):

PR #583, opened 2026-04-23
feat: add BucketLifeTimeILM to ScannerMetrics

krisis:
## Description

SDK consumers can now read per-bucket ILM action counters from `ScannerMetrics`
in addition to the existing cluster-wide counters.

## Motivation and Context

Prerequisite for miniohq/eos#4526. The eos side gates population on a bucket
parameter, so cluster-wide queries do not fan out the payload by bucket count.

## How to test this PR?

```bash
go generate ./...
go test ./... -run "TestScannerMetricsMerge|TestEncodeDecodeBucketILMStats|TestMarshalUnmarshalBucketILMStats|TestEncodeDecodeScannerMetrics" -v
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/estream/types.go
+++ b/estream/types.go
@@ -19,7 +19,7 @@
 
 package estream
 
-//go:generate stringer -type=blockID -trimprefix=block
+//go:generate go tool stringer -type=blockID -trimprefix=block
 
 type blockID int8
 
@@ -36,7 +36,7 @@ const (
 
 type checksumType uint8
 
-//go:generate stringer -type=checksumType -trimprefix=checksumType
+//go:generate go tool stringer -type=checksumType -trimprefix=checksumType
 
 const (
 	checksumTypeNone checksumType = iota
--- a/metrics.go
+++ b/metrics.go
@@ -459,6 +459,34 @@ func (r *Metrics) Merge(other *Metrics) {
 	}
 }
 
+// BucketILMStats reports the cumulative ILM action counters for a single
+// bucket carried in a ScannerMetrics value.
+type BucketILMStats struct {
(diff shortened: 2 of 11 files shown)


estream/types.go
```diff
@@ -19,7 +19,7 @@
 
 package estream
 
-//go:generate stringer -type=blockID -trimprefix=block
+//go:generate go tool stringer -type=blockID -trimprefix=block
```
klauspost, line 22: Oh. you can do that. Neat.

metrics.go
```diff
@@ -454,6 +454,11 @@ type ScannerMetrics struct {
 	// Number of accumulated ILM operations by type since server restart.
 	LifeTimeILM map[string]uint64 `json:"ilm_ops,omitempty"`
 
+	// Number of accumulated ILM operations by type since server start, per bucket.
+	// Keyed by bucket name, then by action name. Only populated when a specific
+	// bucket is requested; nil otherwise.
+	BucketLifeTimeILM map[string]map[string]uint64 `json:"bucket_ilm_ops,omitempty"`
```
klauspost: OK, so you are returning for a specific bucket. You need to add the bucket name in the params.  Should be fairly trivial.

Make it a pointer struct with the bucket name and fields, so we can add more without having to add individual fields here.

krisis:
@klauspost thanks for your advance review. I've addressed your comments. PTAL.

klauspost (approved):

kannappanr (approved):

PR #584, opened 2026-04-24
persist additional replication metrics for bucket/arn

poornas:


prakashsvmx (approved):

reivaj05 (approved):

klauspost:
Seems like you are removing any possibility of making this feasible to call without risking self-nuking both servers and callers.

Is it REALLY needed to have this many dimensions on the data?  It is called a "Summary", but it includes everything and the kitchen sink.

poornas:
> Seems like you are removing any possibility of making this feasible to call without risking self-nuking both servers and callers.
> 
> Is it REALLY needed to have this many dimensions on the data? It is called a "Summary", but it includes everything and the kitchen sink.

the console UI has taken a life of its own - if we are to move away from in-memory stats (used by mc and prometheus) towards persisted stats provided by real time metrics, then all this extra dimensions need to be saved. relying partially on in-memory vs persisted will add further to the confusion

trying to understand why nested maps pose a risk for callers - this is consulted only on startup, AFAIK.

PR #590, opened 2026-05-06
feat: add MetricsBuckets type and BucketAPIMetrics for per-bucket monitoring

0xMALVEE:
feat: add MetricsBuckets type and BucketAPIMetrics for per-bucket monitoring

required for pr: https://github.com/miniohq/eos/pull/4877

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -65,6 +65,7 @@ const (
 	MetricsReplication
 	MetricsProcess
 	MetricsHealing
+	MetricsBuckets
 
 	// MetricsAll must be last.
 	// Enables all metrics.
@@ -101,6 +102,7 @@ func (m MetricType) String() string {
 	addIf(m.Contains(MetricsReplication), "Replication")
 	addIf(m.Contains(MetricsProcess), "Process")
 	addIf(m.Contains(MetricsHealing), "Healing")
+	addIf(m.Contains(MetricsBuckets), "Buckets")
 	return b.String()
 }
 
@@ -159,6 +161,7 @@ type MetricsOptions struct {
 	DrivePoolIdx []int         // Only include metrics for these drive pools. Leave empty for all.
 	DriveSetIdx  []int         // Only include metrics for these drive sets (combine with PoolIdx if needed).
 	Disks        []string      // Include only specific disks. Leave empty for all.
+	Buckets      []string      // Include only specific buckets in bucket metrics. Leave empty for all.
 	ByJobID      string
 	ByDepID      string
 
@@ -195,6 +198,9 @@ func (adm *AdminClient) Metrics(ctx context.Context, o MetricsOptions, out func(
 	}
 
 	q.Set("disks", strings.Join(o.Disks, ","))
+	if len(o.Buckets) > 0 {
+		q.Set("buckets", strings.Join(o.Buckets, ","))
+	}
 	if o.ByDisk {
 		q.Set("by-disk", "true") // Legacy flag
 		o.Flags.Add(MetricsByDisk)
@@ -368,6 +374,7 @@ type Metrics struct {
 	Replication *ReplicationMetrics `json:"replication,omitempty"`
 	Process     *ProcessMetrics     `json:"process,omitempty"`
 	Healing     *HealingMetrics     `json:"healing,omitempty"`
+	Buckets     *BucketAPIMetrics   `json:"buckets,omitempty"`
 }
 
 // Merge other into r.
@@ -435,6 +442,10 @@ func (r *Metrics) Merge(other *Metrics) {
 		r.Healing = &HealingMetrics{}
 	}
 	r.Healing.Merge(other.Healing)
+	if r.Buckets == nil && other.Buckets != nil {
+		r.Buckets = &BucketAPIMetrics{}
+	}
+	r.Buckets.Merge(other.Buckets)
 }
 
 // ScannerMetrics contains scanner information.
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -2960,3 +2967,140 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+	BucketHistory map[string]BucketTimeSeries `json:"bucketHistory,omitempty"`
+
+	// HistorySlots is the number of time-series slots.
+	HistorySlots int `json:"historySlots,omitempty"`
+
+	// HistoryIntervalSecs is the time-series slot duration in seconds.
+	HistoryIntervalSecs int `json:"historyIntervalSecs,omitempty"`
+
+	// HistoryFirstTime is the timestamp of the oldest time-series slot.
+	HistoryFirstTime time.Time `json:"historyFirstTime,omitempty"`
```
klauspost: OK. So these are all by bucket. Let's...

A) Make a top level `Buckets map[string]BucketMetrics`, so we don't have buckets duplicated for all metric types.
B) Add a top level counter `N int ...` - so you know how many nodes reported.
C) Make specific types for different time length segments, so different ones can be requested at the same time.
D) Be sure that intervals are extended to the longest. You may have nodes that report shorter history than others. I think you do this, but please check.
E) Consider adding merge functions to the types instead of doing all on the top element. That will make it easier to aggregate.

Add option to specify buckets on requests. See how this is done with drives for example.

metrics.go
```diff
@@ -2960,3 +2967,140 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+	Requests  int64 `json:"requests"`
+	Errors4xx int64 `json:"errors4xx"`
+	Errors5xx int64 `json:"errors5xx"`
+}
+
+// BucketIOStat holds per-bucket S3 byte counters.
+type BucketIOStat struct {
+	BytesIn  uint64 `json:"bytesIn"`
+	BytesOut uint64 `json:"bytesOut"`
+}
```
klauspost: Any particular reason these are separate?

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+
+// BucketMetrics holds all API stats for one bucket within a single time window:
+// per-operation request totals, byte I/O counters, and a time-series.
+type BucketMetrics struct {
+	// Totals holds per-S3-operation request counts for this bucket.
+	Totals map[string]BucketOpStat `json:"totals,omitempty"`
+
+	// BytesIn and BytesOut are cumulative byte counters for this bucket.
+	BytesIn  uint64 `json:"bytesIn"`
+	BytesOut uint64 `json:"bytesOut"`
```
klauspost: These don't make sense.

A) We don't know how many servers.
B) We don't know the time range.
C) It should be trivial to compute from buckets.

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+	Errors    []int64 `json:"errors"`
+	Errors4xx []int64 `json:"errors4xx,omitempty"`
+	Errors5xx []int64 `json:"errors5xx,omitempty"`
+}
+
+// BucketMetrics holds all API stats for one bucket within a single time window:
+// per-operation request totals, byte I/O counters, and a time-series.
+type BucketMetrics struct {
+	// Totals holds per-S3-operation request counts for this bucket.
+	Totals map[string]BucketOpStat `json:"totals,omitempty"`
```
```suggestion
	Buckets map[string]BucketOpStat `json:"totals,omitempty"`
```

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+
+	// BytesIn and BytesOut are cumulative byte counters for this bucket.
+	BytesIn  uint64 `json:"bytesIn"`
+	BytesOut uint64 `json:"bytesOut"`
+
+	// History holds the time-series. Slots are ordered oldest-first.
+	History BucketTimeSeries `json:"history,omitempty"`
```
klauspost: Add this to the ByBucket, so there is only one map.

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+	N int `json:"n"`
+
+	// Buckets maps bucket name to its consolidated metrics.
+	Buckets map[string]BucketMetrics `json:"buckets,omitempty"`
+
+	// HistoryIntervalSecs is the duration of each time-series slot in seconds.
+	HistoryIntervalSecs int `json:"historyIntervalSecs,omitempty"`
+
+	// HistoryFirstTime is the timestamp of the oldest time-series slot.
+	HistoryFirstTime time.Time `json:"historyFirstTime,omitempty"`
```
klauspost: Allow data to be reported for more than one time resolution similar to other types. Make them pointers, so they can be set or not set.

This means each bucket can have different start times.

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+type BucketOpStat struct {
+	Requests  int64 `json:"requests"`
+	Errors4xx int64 `json:"errors4xx"`
+	Errors5xx int64 `json:"errors5xx"`
+}
+
+// BucketTimeSeries holds pre-aggregated time-series for one bucket.
+// Slots are ordered oldest-first; index len-1 is the most recent.
+type BucketTimeSeries struct {
+	Total     []int64 `json:"total"`
```
klauspost: Total what?  Is it requests? Bytes?

metrics.go
```diff
@@ -2960,3 +2971,150 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+
+	// History holds the time-series. Slots are ordered oldest-first.
+	History BucketTimeSeries `json:"history,omitempty"`
+}
+
+// BucketWindowMetrics holds aggregated bucket metrics for one time window
+// (minute, hour, or day). N is the number of nodes that contributed data.
+type BucketWindowMetrics struct {
+	// N is the number of nodes that reported data for this window.
+	N int `json:"n"`
```
klauspost: Consider moving this to the top level.

klauspost (changes requested):

metrics.go
```diff
@@ -2960,3 +2971,158 @@ func (m *HealingMetrics) Merge(other *HealingMetrics) {
...
+
+	// HistoryIntervalSecs is the duration of each time-series slot in
+	// seconds. Per-bucket so heterogeneous nodes reporting different
+	// resolutions can each contribute. nil when the bucket has no
+	// time-series.
+	HistoryIntervalSecs *int `json:"historyIntervalSecs,omitempty"`
+
+	// HistoryFirstTime is the timestamp of the oldest time-series slot.
+	// Per-bucket so each bucket can have a distinct start time.
+	HistoryFirstTime *time.Time `json:"historyFirstTime,omitempty"`
```
klauspost: OK. 

* Add LastMinute. Aggregated, meaning not time segmented. Always present.
* Add LastHour, populated if flag is set, segmented, 1m.
* Add LastDay, populated if flag set, segmented in 15m.

0xMALVEE: @klauspost I wanted lastminute also be time segmented, so we can show the charts for 1 minutes also.

0xMALVEE:
@klauspost take a look

klauspost (approved):
lgtm! Gives valuable data for multiple use cases and should still be reasonably sized - as well as give good caller flexibility.

metrics.go
klauspost: No. That will just be too noisy. Showing the last minute average is a good enough for "current speed".

klauspost: You are striking a good balance by having it separated by API, but only aggregate. If you do by second, you'd have 60x the data with little added value.

0xMALVEE: ok removed the segmented lastMinute, only aggregating.

shtripat (approved):
lgtm

PR #592, opened 2026-05-07
feat(scanner): add ExcessiveVersionObjects to ScannerMetrics

shtripat:
## Description

`ScannerMetrics` already exposes `ExcessivePrefixes` (folders with too
many sub-entries) via the realtime metrics API, which collects per-node
reports and merges them cluster-wide. This PR does the same for the
version-excess side: add `ExcessiveVersionObjects []string` so objects
exceeding the version count or cumulative size threshold are visible in
`mc admin top scanner` and Prometheus scrapes across all nodes.

A `DiscardedExcessEntries uint64` counter tracks entries dropped when
the merged list is capped at 100 entries, letting operators know the
list is incomplete without polluting the payload.

## Changes

- **`ScannerMetrics`**: two new fields
  - `ExcessiveVersionObjects []string` (`json:"excessive_versions"`)
  - `DiscardedExcessEntries uint64` (`json:"discarded_excess_entries"`)
- **`ScannerMetrics.Merge()`**: replaces the unbounded
  `ExcessivePrefixes` dedup map with a `setof.String`-based helper
  that caps both excess slices at 100 entries and accumulates
  `DiscardedExcessEntries`
- **`metrics_gen.go`**: regenerated

## Testing

New test cases in `TestScannerMetricsMerge`:
- `ExcessiveVersionObjects` dedup and sort
- 100-entry cap with correct `DiscardedExcessEntries` count (110
  unique → 100 kept, 10 discarded)
- Accumulated `DiscardedExcessEntries` across successive merges

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -38,7 +38,6 @@ import (
 	"github.com/shirou/gopsutil/v4/cpu"
 	"github.com/shirou/gopsutil/v4/load"
 	"github.com/tinylib/msgp/msgp"
-	"github.com/tinylib/msgp/msgp/setof"
 )
 
 //go:generate msgp -unexported -d clearomitted -d "tag json" -d "timezone utc" -d "maps binkeys" -file $GOFILE
@@ -471,13 +470,12 @@ type ScannerMetrics struct {
 
 	// ExcessivePrefixes lists prefixes marked as having excessive sub-entries
 	// within the last 24 hours.
-	// Capped at 100 entries per cross-node merge; see DiscardedExcessEntries.
-	ExcessivePrefixes setof.String `json:"excessive,omitempty"`
+	ExcessivePrefixes []string `json:"excessive,omitempty"`
 
 	// ExcessiveVersionObjects lists objects that have exceeded the version
 	// count or cumulative size threshold within the last 24 hours.
 	// Capped at 100 entries per cross-node merge; see DiscardedExcessEntries.
-	ExcessiveVersionObjects setof.String `json:"excessive_versions,omitempty"`
+	ExcessiveVersionObjects []string `json:"excessive_versions,omitempty"`
 
 	// DiscardedExcessEntries counts entries dropped beyond the 100-entry cap
 	// during cross-node merge. This counter is not deduplicated.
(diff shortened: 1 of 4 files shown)


metrics.go
```diff
@@ -468,10 +469,20 @@ type ScannerMetrics struct {
...
-	// Paths that have been marked as having excessive number of entries within the last 24 hours.
+	// ExcessivePrefixes lists prefixes marked as having excessive sub-entries
+	// within the last 24 hours.
+	// Capped at 100 entries per cross-node merge; see DiscardedExcessEntries.
 	ExcessivePrefixes []string `json:"excessive,omitempty"`
 
+	// ExcessiveVersionObjects lists objects that have exceeded the version
+	// count or cumulative size threshold within the last 24 hours.
+	// Capped at 100 entries per cross-node merge; see DiscardedExcessEntries.
+	ExcessiveVersionObjects []string `json:"excessive_versions,omitempty"`
```
klauspost: I meant make this `setof.String`

```suggestion
	ExcessiveVersionObjects setof.String `json:"excessive_versions,omitempty"`
```

So only when you have > 100 after a merge you need to sort+truncate. Makes the default path much faster.

You *can* change `ExcessivePrefixes` above to the same (setof is stored like an array). Actually you can do that without really breaking anything. Not seeing anything that really depends on it, so maybe consider that as well.

shtripat: ah ok. Let me do that. Thanks

klauspost (approved):
lgtm

metrics.go
```diff
@@ -468,9 +469,19 @@ type ScannerMetrics struct {
 	// Currently active path(s) being scanned.
 	ActivePaths []string `json:"active,omitempty"`
 
-	// Excessive prefixes.
-	// Paths that have been marked as having excessive number of entries within the last 24 hours.
-	ExcessivePrefixes []string `json:"excessive,omitempty"`
+	// ExcessivePrefixes lists prefixes marked as having excessive sub-entries
+	// within the last 24 hours.
+	// Capped at 100 entries per cross-node merge; see DiscardedExcessEntries.
+	ExcessivePrefixes setof.String `json:"excessive,omitempty"`
```
Praveenrajmani: Isn't this a breaking change? JSON wire format was `excessive":["a","b"]`, now `excessive":{"a":{},"b":{}}`; can you verify if this doesn't break any JSON consumers @shtripat

Praveenrajmani: https://github.com/tinylib/msgp/blob/master/msgp/setof/generated.go has no MarshalJSON/UnmarshalJSON

klauspost: For JSON, yes.  Oh yeah, we do use it for mc - forgot about that... I will propose adding it to tinylib. 

Let's revert the change - but keep the new one - for now.

shtripat: ok, will update the PR. Thanks

klauspost: Suggested https://github.com/tinylib/msgp/pull/443

PR #595, opened 2026-05-12
feat(health): add HealthDataTypeSysProductInfo

grahit13:
## Summary                                                                                                                                                 
                                                                                                                                                           
  Adds `HealthDataTypeSysProductInfo` (`"sysproductinfo"`) plus matching                                                                                     
  map/list entries. Purely additive — no struct or API changes.
                                                                                                                                                             
  ## Motivation                                                                                                                                            

  `SysInfo.ProductInfo`, `GetProductInfo`, and `getDMIInfo` already exist                                                                                    
  but had no `HealthDataType`, so the section was impossible to request
  via `ServerHealthInfo`. Server-side wiring lands in the matching                                                                                           
  eos PR.    
  Fixes: https://github.com/miniohq/subnet/issues/6318
  
 ## How to test?
  - Build minio from miniohq/eos#5047                                                                                                                      
  - Build mc from miniohq/ec#509
  - `mc support diag --airgap <alias>`
  - `gunzip -c diag-data-*.json.gz | jq -s 'last | .sys.productinfo'`
  - Expect one entry per node: real vendor/name/version on Linux with DMI, `"unknown"` fallback elsewhere                                                    
  - Re-run with `--anonymize strict`: `serial_number`/`uuid` blanked, others preserved  
  
  ## Expected Output(standard mode, on a Linux host with DMI)
                                                                                                                                                             
  ```json                                                                                                                                                  
  [
    {
      "addr": "node-1:9000",
      "family": "PowerEdge",                                                                                                                                 
      "name": "PowerEdge R750",
      "vendor": "Dell Inc.",                                                                                                                                 
      "serial_number": "5XYZ123",                                                                                                                          
      "uuid": "4c4c4544-0058-5910-8050-cac04f445232",                                                                                                        
      "sku": "SKU=PE-R750",                                                                                                                                  
      "version": "2.10.0"                                                                                                                                    
    }                                                                                                                                                        
  ]

PR #597, opened 2026-05-13
fix(health): omit empty DMI fields from ProductInfo

grahit13:
## Summary                                                                                                                                                 
                                                                                                                                                           
  - `getDMIInfo` returns `""` instead of the `"unknown"` sentinel on read error.                                                                             
  - All `ProductInfo` JSON tags marked `,omitempty` so unfetched fields drop out.
                                                                                                                                                             
  ## Motivation                                                                                                                                            
                                                                                                                                                             
  Follow-up to #595. On non-Linux hosts and on non-root Linux MinIO (the                                                                                     
  majority of deployments, since the kernel restricts `product_serial` /
  `product_uuid` to root), the current output includes `"field": "unknown"`                                                                                  
  for every unreadable DMI file — noisy and not useful.                                                                                                    
                                                                                                                                                             
  After this PR, only fields with real values appear in the JSON. Sparse                                                                                     
  entries (e.g., a macOS dev host) collapse to just `{"addr": "..."}`.

### DIFF
--- a/health.go
+++ b/health.go
@@ -794,19 +794,19 @@ func getXFSErrorMaxRetries() XFSErrorConfigs {
 type ProductInfo struct {
 	NodeCommon
 
-	Family       string `json:"family"`
-	Name         string `json:"name"`
-	Vendor       string `json:"vendor"`
-	SerialNumber string `json:"serial_number"`
-	UUID         string `json:"uuid"`
-	SKU          string `json:"sku"`
-	Version      string `json:"version"`
+	Family       string `json:"family,omitempty"`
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

harshavardhana:
I am assuming healthanalyzer on subnet side can handle `omitempty` missing fields.

shtripat (approved):
lgtm

anjalshireesh (approved):

grahit13:
> I am assuming healthanalyzer on subnet side can handle `omitempty` missing fields.

Yes @harshavardhana

PR #600, opened 2026-05-13
audit log: introduce message + metadata format, deprecate AuditDetails

Praveenrajmani:
Add a message + flexible-metadata shape to `log.Audit` so producers can describe events in plain English with arbitrary
structured payloads instead of populating a fixed typed substruct.

Reading an audit entry previously meant: route on Category → pick the matching arm of the AuditDetails union → stringify its fields. Adding a new audit subject required a new substruct + a new wrapper in the producer + msgp regen. The new shape lets the producer ship a short, human-readable line and any structured fields it wants, with no schema bumps.

Companion change in eos rewires every audit producer to populate Message/Metadata; this PR is the wire-format and
type-surface piece.

### DIFF
--- a/log/audit.go
+++ b/log/audit.go
@@ -66,11 +66,16 @@ const (
 	AuditCategoryBatch          AuditCategory = "batch"
 )
 
-// AuditAction represents the type of action performed
+// AuditAction represents the type of action performed.
+//
+// Deprecated: no longer set by the audit recorder. Kept for
+// wire-format compatibility.
 //
 //msgp:shim AuditAction as:string
 type AuditAction string
 
+// Deprecated: no longer set by the audit recorder. Kept for
+// wire-format compatibility.
 const (
 	AuditActionCreate  AuditAction = "create"
 	AuditActionUpdate  AuditAction = "update"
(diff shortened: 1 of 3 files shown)


log/helper.go
```diff
@@ -24,6 +24,18 @@ import (
...
 )
 
+// MetadataKeyDiff is the canonical metadata key under which Diff entries
+// appear in audit log Metadata.
+const MetadataKeyDiff = "diff"
+
+// Diff carries the before/after state of a textual config payload for audit
+// log metadata. Consumers compose a +/- diff (or any other view) from the
+// Old/New fields on demand.
+type Diff struct {
```
shtripat, line 34: is it like consumer can use strings like `enable=on endpoint=http://127.0.0.1:8000/webhook batch_size=1000` ?

shtripat (approved):
lgtm, one question

log/helper.go
Praveenrajmani: yes, producer will just set the old and new items; the consumers can use this to find the Diff() with some helper packages for example -  https://github.com/sergi/go-diff

for now, mc is the only consumer and it just wants raw JSON in --json mode

log/audit.go
```diff
@@ -408,5 +460,5 @@ func (t *TierAuditDetails) Redact() {}
 
 // String returns a simple string representation for Audit (required by eos LogEntry interface)
 func (a Audit) String() string {
-	return fmt.Sprintf("audit: category=%s action=%s api=%s", a.Category, a.Action, a.APIName)
+	return fmt.Sprintf("audit: category=%s api=%s message=%s", a.Category, a.APIName, a.Message)
```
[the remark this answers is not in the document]
Praveenrajmani: no need this

grahit13 (approved):
LGTM, verified!

harshavardhana (approved):

PR #602, opened 2026-05-15
feat: add multi-value filter fields to APILogOpts

csumisha:
Add StatusCodes, Origins, Types, and BucketPrefix to APILogOpts to support filtering API logs by multiple values and bucket name prefix.

### DIFF
--- a/api-logs.go
+++ b/api-logs.go
@@ -39,12 +39,12 @@ type APILogOpts struct {
 	Bucket       string        `json:"bucket,omitempty"`
 	BucketPrefix string        `json:"bucketPrefix,omitempty"`
 	Prefix       string        `json:"prefix,omitempty"`
-	StatusCode   int           `json:"statusCode,omitempty"`  // Deprecated: use StatusCodes
+	StatusCode   int           `json:"statusCode,omitempty"` // Deprecated: use StatusCodes
 	StatusCodes  []string      `json:"statusCodes,omitempty"`
 	Interval     time.Duration `json:"interval,omitempty"`
-	Origin       log.Origin    `json:"origin,omitempty"`  // Deprecated: use Origins
+	Origin       log.Origin    `json:"origin,omitempty"` // Deprecated: use Origins
 	Origins      []string      `json:"origins,omitempty"`
-	Type         log.APIType   `json:"type,omitempty"`  // Deprecated: use Types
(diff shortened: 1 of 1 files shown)


Praveenrajmani (approved):

shtripat (approved):
lgtm

grahit13:
Please check https://github.com/minio/madmin-go/actions/runs/25907566539/job/76144028003?pr=602

grahit13 (approved):

api-logs.go
```diff
@@ -34,16 +34,20 @@ import (
...
-	Type       log.APIType   `json:"type,omitempty"`
-	MaxPerNode int           `json:"maxPerNode,omitempty"` // Deprecated
-	Limit      int           `json:"limit,omitempty"`
+	Node         string        `json:"node,omitempty"`
+	API          string        `json:"api,omitempty"`
+	Bucket       string        `json:"bucket,omitempty"`
+	BucketPrefix string        `json:"bucketPrefix,omitempty"`
+	Prefix       string        `json:"prefix,omitempty"`
+	StatusCode   int           `json:"statusCode,omitempty"` // Deprecated: use StatusCodes
+	StatusCodes  []string      `json:"statusCodes,omitempty"`
```
klauspost: Why is this a string?

api-logs.go
```diff
@@ -34,16 +34,20 @@ import (
...
+	Interval     time.Duration `json:"interval,omitempty"`
+	Origin       log.Origin    `json:"origin,omitempty"` // Deprecated: use Origins
+	Origins      []string      `json:"origins,omitempty"`
```
klauspost: Type-less seems strictly worse. It is not really clear what these are.

api-logs.go
csumisha: Kept it for 4xx or 5xx kind of search but will keep it as int and expand it before sending it.

klauspost: I see. Not against that per se. It does need some explanation/docs here if you go that route.

klauspost: It is more that a string can be anything, so from an API standpoint it needs to be clearer.

csumisha: For clairty, separated statusCodes as int and added statusRanges as string.

klauspost (approved):

PR #603, opened 2026-05-15
Rename inventory to "system inventory"

rraulinio:
Rename `inventory` to `system inventory` in order to make the difference clear.


shtripat (approved):
lgtm, I understand this is not yet released so rename is fine

klauspost (approved):
(title need quotes ;)

PR #604, opened 2026-05-16
Added multiple search support for api names and bucket names

csumisha:
Add APIs []string and Buckets []string to APILogOpts for multi-value  filtering. Both support case-insensitive matching: entries ending with "*" are prefix matches, otherwise substring match.

### DIFF
--- a/api-logs.go
+++ b/api-logs.go
@@ -34,18 +34,14 @@ import (
 
 // APILogOpts represents the options for the APILogOpts
 type APILogOpts struct {
-	Node         string        `json:"node,omitempty"`
-	API          string        `json:"api,omitempty"`
-	Bucket       string        `json:"bucket,omitempty"`
-	BucketPrefix string        `json:"bucketPrefix,omitempty"`
+	Nodes        []string      `json:"nodes,omitempty"`   // entries ending with "*" are prefix matches, otherwise exact match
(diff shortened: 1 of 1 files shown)


klauspost (commented):
Honestly, we should just add a v2 API.

You can't send new values in a backwards compatible way anyway. So just make a new endpoint.

csumisha:
> Honestly, we should just add a v2 API.
> 
> You can't send new values in a backwards compatible way anyway. So just make a new endpoint.

@klauspost my bad these apis are not live yet, so no need of deprecated keys. I will remove it.

shtripat (approved):
lgtm, @klauspost, it makes sense I feel now, right?

grahit13 (approved):

klauspost:
@csumisha Yeah, wasn't sure either, and GIT said the changes were quite old, so I just assumed it was live.

klauspost (approved):
Please have EOS changes ready for a quick merge, so we wont have this too long in master without having it in EOS.

(otherwise mark this as "draft")

csumisha:
> Please have EOS changes ready for a quick merge, so we wont have this too long in master without having it in EOS.
> 
> (otherwise mark this as "draft")

Thanks @klauspost both EOS and EC changes are ready.

PR #607, opened 2026-05-19
feat: add WithIOStats counter to DiskIOStats

anjalshireesh:
Adds a `WithIOStats int` field to `DiskIOStats` that counts the drives whose
block-device IO statistics were available (i.e. the kernel sysfs `stat` file
could be read). Network-attached storage and most k8s setups don't expose
this file, so iostat fields stay zero — the new counter lets consumers tell
"no iostat data on this setup" apart from "no I/O happened" and decide
whether to render iostat columns/charts.

Living inside `DiskIOStats` means `DiskIOStats.Add` merges it for free, no
custom logic in `DiskMetric.Merge`. After aggregation, `with_iostats` is
the count of contributing drives — strictly more informative than an
OR-merged bool that would say "available" even if only 1 of N drives
reported.

`DiskIOStatsLegacy` mirrors the field to preserve the direct Go type
conversions used elsewhere in `DiskMetric.Merge`; with `omitempty` the
legacy wire format is unaffected when the value is zero.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -704,7 +704,10 @@ func (s *ScannerMetrics) Merge(other *ScannerMetrics) {
 
 // DiskIOStats contains IO stats of a single drive
 type DiskIOStats struct {
-	N              int    `json:"n,omitempty"`
+	N int `json:"n,omitempty"`
+	// WithIOStats is the subset of N whose kernel sysfs iostat
+	// (`/sys/dev/block/.../stat`) was readable; consumers gate I/O UI on it.
+	WithIOStats    int    `json:"with_iostats,omitempty"`
 	ReadIOs        uint64 `json:"read_ios,omitempty"`
 	ReadMerges     uint64 `json:"read_merges,omitempty"`
 	ReadSectors    uint64 `json:"read_sectors,omitempty"`
@@ -726,8 +729,11 @@ type DiskIOStats struct {
 	BitrotHealed   uint64 `json:"bitrot_healed,omitempty"`
 }
 
+// DiskIOStatsLegacy mirrors DiskIOStats field-for-field so direct Go type
+// conversions (used in DiskMetric.Merge) stay valid; mirror new fields too.
 type DiskIOStatsLegacy struct {
 	N              int    `json:"n,omitempty"`
+	WithIOStats    int    `json:"with_iostats,omitempty"`
 	ReadIOs        uint64 `json:"read_ios,omitempty"`
 	ReadMerges     uint64 `json:"read_merges,omitempty"`
 	ReadSectors    uint64 `json:"read_sectors,omitempty"`
@@ -755,6 +761,7 @@ func (d *DiskIOStats) Add(other *DiskIOStats) {
 		return
 	}
 	d.N += other.N
+	d.WithIOStats += other.WithIOStats
 	d.ReadIOs += other.ReadIOs
 	d.ReadMerges += other.ReadMerges
 	d.ReadSectors += other.ReadSectors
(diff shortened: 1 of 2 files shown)


metrics.go
```diff
@@ -846,6 +846,11 @@ type DiskMetric struct {
 	// Rolling window hourly IO stats (1-minute segments).
 	IOStatsHour SegmentedDiskIO `json:"io_hour"`
 
+	// IOStatsAvailable reports whether block-device IO statistics are
+	// available for this drive. False for network-attached storage and
+	// most k8s setups where the kernel sysfs stat file does not exist.
+	IOStatsAvailable bool `json:"io_stats_available,omitempty"`
```
klauspost: First of all this doesn't merge.

Second why isn't this with the metrics?

klauspost (changes requested):

metrics.go
anjalshireesh: @klauspost is the following approach acceptable?

```
turn the boolean into a counter inside DiskIOStats, e.g.:

  type DiskIOStats struct {
      N                int    `json:"n,omitempty"`            // existing: drives reporting
      NWithIOStats     int    `json:"n_with_iostats,omitempty"` // new: subset where /proc/diskstats was readable
      ReadIOs          uint64 `json:"read_ios,omitempty"`
      ...
  }

  This way:
  - Add() in DiskIOStats already sums fields — NWithIOStats gets merged for free, no DiskMetric.Merge() change needed.
  - The aggregate carries useful information: "N drives report, K of them have IO stats." UI can decide threshold (e.g., render only if K == N, or K > 0).
  - The flag is co-located with the IO stats it describes

  Eos-side, the collector would do s.NWithIOStats = 1 when seeding succeeds (and 0 otherwise), instead of toggling the parent-level boolean.
```

anjalshireesh: My original intention was to capture simply whether IOStats data can be fetched in this setup. If we can detect that and capture at a higher level, we don't need to look into each drive level metrics when iostats is not available. The UI can simply check the value of this bool and hide all relevant columns/charts when false.

Would appreciate your guidance on the best way to achieve this @klauspost

klauspost: Yeah, just name it `WithIOStats`. I guess originally it wouldn't increment N if it wasn't, but now it will, with bitrot stuff being added.

Existing use of 'N' will need to be updated, so it will work backwards compatible where servers aren't sending `WithIOStats`. Either that or count "bitrot" divisor separately - `WithBitrot int` - probably that is easier, though it would also need fallback, if it is used anywhere.

anjalshireesh: Modified to use `WithIOStats`. PTAL @klauspost

harshavardhana (approved):

anjalshireesh:
@klauspost PTAL

metrics.go
```diff
@@ -726,8 +731,15 @@ type DiskIOStats struct {
 	BitrotHealed   uint64 `json:"bitrot_healed,omitempty"`
 }
 
+// DiskIOStatsLegacy mirrors DiskIOStats field-for-field so callers can
+// use direct Go type conversions (DiskIOStats(legacy), DiskIOStatsLegacy(cur))
+// instead of writing field-by-field copy code at every call site, including
+// in DiskMetric.Merge below. New fields added to DiskIOStats should be
+// mirrored here to preserve that property; with `omitempty` JSON tags,
+// legacy wire payloads are unaffected when the value is zero.
```
rraulinio: Please shave this.

anjalshireesh: updated with a more concise comment

metrics.go
```diff
@@ -704,7 +704,12 @@ func (s *ScannerMetrics) Merge(other *ScannerMetrics) {
 
 // DiskIOStats contains IO stats of a single drive
 type DiskIOStats struct {
-	N              int    `json:"n,omitempty"`
+	N int `json:"n,omitempty"`
+	// WithIOStats is the subset of N for which block-device IO statistics
+	// were available (i.e. /sys/dev/block/<major>:<minor>/stat could be
+	// read). When merged across drives this gives the count of contributing
+	// drives, letting consumers decide whether to render the I/O surface.
```
rraulinio: This could also be compacted.

rraulinio (commented):
Looks good, one nit 'tho.

rraulinio (approved):

klauspost (approved):
Yeah, shorten the essay, otherwise lgtm

metrics.go
anjalshireesh: done

PR #608, opened 2026-05-21
feat: add multi-value filter fields to AuditLogOpts

grahit13:
Replace Node/API/Bucket/Category singletons on AuditLogOpts with
plural slice fields (Nodes/APIs/Buckets/Categories); within-field
OR, cross-field AND. MaxPerNode deprecated in favor of Limit.                                                                            
Mirrors the APILogOpts shape from #604.                                                                                                  
                                                                                                                                          
Wildcard syntax on Nodes/APIs/Buckets — for both AuditLogOpts                                                                            
and APILogOpts — now supports abc*, *abc, *abc*, and *  
(case-insensitive); bare values are exact matches.                                                                                                        
   
  Paired with [eos](https://github.com/miniohq/eos/pull/5254) + [ec](https://github.com/miniohq/ec/pull/523) PRs that consume the new shape.  
  
  Ref: https://github.com/miniohq/eos/issues/4967

### DIFF
--- a/audit-logs.go
+++ b/audit-logs.go
@@ -34,12 +34,13 @@ import (
 
 // AuditLogOpts represents the options for the audit logs
 type AuditLogOpts struct {
-	Node       string            `json:"node,omitempty"`
-	API        string            `json:"api,omitempty"`
-	Bucket     string            `json:"bucket,omitempty"`
-	Interval   time.Duration     `json:"interval,omitempty"`
-	Category   log.AuditCategory `json:"category,omitempty"`
-	MaxPerNode int               `json:"maxPerNode,omitempty"`
+	Nodes      []string            `json:"nodes,omitempty"`   // entries ending with "*" are prefix matches, otherwise exact match
+	APIs       []string            `json:"apis,omitempty"`    // entries ending with "*" are prefix matches, otherwise exact match
+	Buckets    []string            `json:"buckets,omitempty"` // entries ending with "*" are prefix matches, otherwise exact match
+	Interval   time.Duration       `json:"interval,omitempty"`
+	Categories []log.AuditCategory `json:"categories,omitempty"`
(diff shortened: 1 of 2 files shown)


audit-logs.go
```diff
@@ -34,12 +34,13 @@ import (
...
 type AuditLogOpts struct {
-	Node       string            `json:"node,omitempty"`
-	API        string            `json:"api,omitempty"`
-	Bucket     string            `json:"bucket,omitempty"`
-	Interval   time.Duration     `json:"interval,omitempty"`
-	Category   log.AuditCategory `json:"category,omitempty"`
-	MaxPerNode int               `json:"maxPerNode,omitempty"`
+	Nodes      []string            `json:"nodes,omitempty"`   // entries ending with "*" are prefix matches, otherwise exact match
+	APIs       []string            `json:"apis,omitempty"`    // entries ending with "*" are prefix matches, otherwise exact match
+	Buckets    []string            `json:"buckets,omitempty"` // entries ending with "*" are prefix matches, otherwise exact match
```
klauspost: What about `*xyz` -> suffix match. `*xyz*` -> contains?

Seems annoying not to have a "contains".

grahit13: Done @klauspost  — extended the wildcards to support abc* (prefix), *abc (suffix), *abc* (contains), and * (any) on both `AuditLogOpts` and `APILogOpts`. Since the eos PR refactors API and audit logs to share filter primitives, it made sense to update both shapes together; matching eos + mc changes will land alongside.
 
 CC @csumisha 
 
 `ErrorLogOpts` still uses the singular Node/Bucket/API fields — we'll bring it onto the same multi-value + wildcard shape in a follow-up. CC @bexsoft

klauspost (approved):
lgtm

csumisha (approved):

shtripat (approved):

PR #609, opened 2026-05-26
feat: add multi-value filter fields to ErrorLogOpts

reivaj05:
Replace singular node/api/bucket string fields with typed slices supporting wildcard syntax. Prefixes field added for object-key prefix filtering. Mirrors AuditLogOpts (PR #608).

Wildcard rules (case-insensitive within a field, OR-combined):
  "xyz"   → exact match
  "xyz*"  → prefix match
  "*xyz"  → suffix match
  "*xyz*" → contains match
  "*"     → matches anything

part of: https://github.com/miniohq/eos/issues/5312

### DIFF
--- a/error-logs.go
+++ b/error-logs.go
@@ -50,6 +50,17 @@ type ErrorLogOpts struct {
 	Prefixes []string      `json:"prefixes,omitempty"`
 	Interval time.Duration `json:"interval,omitempty"`
 	Limit    int           `json:"limit,omitempty"`
+
+	// Deprecated: use Nodes.
+	Node string `json:"node,omitempty"`
+	// Deprecated: use APIs.
+	API string `json:"api,omitempty"`
+	// Deprecated: use Buckets.
+	Bucket string `json:"bucket,omitempty"`
+	// Deprecated: use Prefixes.
+	Prefix string `json:"prefix,omitempty"`
+	// Deprecated: use Limit.
+	MaxPerNode int `json:"maxPerNode,omitempty"`
 }
 
 // GetErrorLogs fetches the persisted error logs from MinIO
--- a/error-logs.go
+++ b/error-logs.go
@@ -32,14 +32,24 @@ import (
 	"github.com/tinylib/msgp/msgp"
 )
 
-// ErrorLogOpts represents the options for the ErrorLogs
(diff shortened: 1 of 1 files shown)


error-logs.go
```diff
@@ -32,14 +32,24 @@ import (
...
-	Prefix     string        `json:"prefix,omitempty"`
-	Interval   time.Duration `json:"interval,omitempty"`
-	MaxPerNode int           `json:"maxPerNode,omitempty"`
+	Nodes    []string      `json:"nodes,omitempty"`
+	APIs     []string      `json:"apis,omitempty"`
+	Buckets  []string      `json:"buckets,omitempty"`
+	Prefixes []string      `json:"prefixes,omitempty"`
+	Interval time.Duration `json:"interval,omitempty"`
+	Limit    int           `json:"limit,omitempty"`
 }
```
harshavardhana, line 65: we cannot add breaking changes like this, please preserve the previous values and use this as a v2 

we should have the right practice of when the structs are designed we cater for future growth as well. 

Can you explain why are we breaking this struct?

Praveenrajmani: +1 if this feature is released and being used by customers; we can mark the existing fields as deprecated

reivaj05: No specific reason, i was just mirroring  the same as in https://github.com/minio/madmin-go/pull/608/changes#diff-b3e07aec6fff9bb353bac83ec300f528caac55ba5737bbf166a445c392573d94L36

I'll add the fields back as deprecated

reivaj05:
@harshavardhana I added back the fields and marked them as deprecated

error-logs.go
grahit13: @reivaj05, PTAL: https://github.com/minio/madmin-go/pull/604#pullrequestreview-4305437943

reivaj05: @grahit13 this is already live, it was introduced last year in https://github.com/miniohq/eos/pull/2173 and edge release already shipped it `EDGE.2026-03-03T18-23-09Z`, we should have deprecated them for the other logs

grahit13: ok my bad, will update accordingly, thanks @reivaj05 
CC @csumisha

grahit13 (approved):

kaankabalak (approved):

PR #613, opened 2026-06-01
Remove redundant forecast fields and keep growth in float64

cniackz:
RecentGrowthRatePerDay and RecentDaysUntilFull duplicated GrowthBytesPerDay
and DaysUntil100Pct: the Kalman filter is already recency-weighted, so the
"Recent" variants always carried the same value as the primary fields. Drop
them.

Change GrowthBytesPerDay from int64 to float64 so the growth rate keeps full
precision instead of truncating a sub-byte-per-day slope to zero.

Document that Variance is the population variance (divided by n). The
capacity-forecast endpoint that fills these is not shipped yet.


shtripat (approved):
I understand as its not released, this removal is fine and only using code PR from eos still work in progress.

PR #614, opened 2026-06-01
Add MetricsTables for Table IO stats

SSushmitha8:
This PR adds support for tracking which Iceberg tables are getting the most traffic across the cluster — reads, writes, and bytes transferred over three time windows: last minute, last hour, and last 24 hours.

  A new metric flag (MetricsTableWindowed) so a node requesting metrics from its peers can ask specifically for table traffic data

```
  TableIOStat      — raw read/write/byte counters for one time window
  TableIOMetrics   — one table's traffic: warehouse + LastMinute/LastHour/LastDay
  TableAPIMetrics  — all active tables keyed by UUID; 
```

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -67,6 +67,7 @@ const (
 	MetricsHealing
 	MetricsBuckets
 	MetricsKMS
+	MetricsTables
 
 	// MetricsAll must be last.
 	// Enables all metrics.
@@ -105,6 +106,7 @@ func (m MetricType) String() string {
 	addIf(m.Contains(MetricsHealing), "Healing")
 	addIf(m.Contains(MetricsBuckets), "Buckets")
 	addIf(m.Contains(MetricsKMS), "KMS")
+	addIf(m.Contains(MetricsTables), "Tables")
 	return b.String()
 }
 
@@ -380,6 +382,7 @@ type Metrics struct {
 	Healing     *HealingMetrics     `json:"healing,omitempty"`
 	Buckets     *BucketAPIMetrics   `json:"buckets,omitempty"`
 	KMS         *KMSRtMetrics       `json:"kms,omitempty"`
+	Tables      *TableAPIMetrics    `json:"tables,omitempty"`
 }
 
 // Merge other into r.
@@ -456,6 +459,12 @@ func (r *Metrics) Merge(other *Metrics) {
 			r.KMS = &KMSRtMetrics{}
 		}
 		r.KMS.Merge(other.KMS)
+	}
+	if other.Tables != nil {
+		if r.Tables == nil {
+			r.Tables = &TableAPIMetrics{}
+		}
+		r.Tables.Merge(other.Tables)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


metrics.go
```diff
@@ -67,6 +67,7 @@ const (
 	MetricsHealing
 	MetricsBuckets
 	MetricsKMS
+	MetricsTableWindowed
```
klauspost: I would like to keep "Tables" for overall table metrics. "Windowed" is a property, not really relevant.

That is why I proposed `MetricsTopTables` to say A) It is about tables and B) To separate it from future table stats, which will be more general than this.


```suggestion
	MetricsTopTables
```

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+}
+
+// TableAPIMetrics holds per-table windowed traffic metrics aggregated
+// across all nodes. Keyed by table UUID.
+type TableAPIMetrics struct {
+	// N is the number of nodes that reported data.
+	N int `json:"n" msg:"n"`
+
+	// Tables maps table UUID to its consolidated windowed metrics.
+	Tables map[string]TableMetrics `json:"tables,omitempty" msg:"tables,omitempty"`
```
klauspost: I am not a particular fan of being string indexed map. It makes merging a tiny bit easier, but use harder.

I feel like a are doing a sorted slice anyway when figuring out the top.

```suggestion
	Top []TableIOMetrics `json:"top,omitempty"`
```

(msgp uses the json tags here, so you can omit them)

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+		if m.LastDay == nil {
+			m.LastDay = &TableTrafficMetrics{}
+		}
+		m.LastDay.Add(other.LastDay)
+	}
+}
+
+// TableAPIMetrics holds per-table windowed traffic metrics aggregated
+// across all nodes. Keyed by table UUID.
+type TableAPIMetrics struct {
```
```suggestion
type TopTableIOMetrics struct {
```

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
 	}
 }
 
+// TableTrafficMetrics holds read/write counts and byte totals for a single
+// table over one time window.
+type TableTrafficMetrics struct {
```
```suggestion
type TableIOMetrics struct {
```

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+	Reads    int64 `json:"reads"    msg:"r"`
```
klauspost, line 3349: Wouldn't we want a nice printable name?
```suggestion
	Name string `json:"name"`
	Reads    int64 `json:"reads"    msg:"r"`
```

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+
+// IsZero reports whether all counters are zero.
+func (t *TableTrafficMetrics) IsZero() bool {
+	return t.Reads == 0 && t.Writes == 0 && t.BytesIn == 0 && t.BytesOut == 0
+}
+
+// TableMetrics holds windowed traffic for a single table.
+type TableMetrics struct {
+	// Warehouse is the warehouse this table belongs to.
+	Warehouse string `json:"warehouse,omitempty" msg:"wh,omitempty"`
```
klauspost: Which "table"? There are many tables in here.

Add it to the tables if you feel it is needed.

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+		}
+		m.LastDay.Add(other.LastDay)
+	}
+}
+
+// TableAPIMetrics holds per-table windowed traffic metrics aggregated
+// across all nodes. Keyed by table UUID.
+type TableAPIMetrics struct {
+	// N is the number of nodes that reported data.
+	N int `json:"n" msg:"n"`
```
klauspost: This belongs on the top level.

metrics.go
klauspost: Maybe UUID - but not sure that is really needed.

tier-azure.go
```diff
@@ -25,22 +25,22 @@ import "errors"
...
-	Endpoint     string `json:",omitempty"`
-	AccountName  string `json:",omitempty"`
-	AccountKey   string `json:",omitempty"`
-	Bucket       string `json:",omitempty"`
-	Prefix       string `json:",omitempty"`
-	Region       string `json:",omitempty"`
-	StorageClass string `json:",omitempty"`
-
-	SPAuth ServicePrincipalAuth `json:",omitempty"`
+	Endpoint     string `json:"endpoint,omitempty"`
```
klauspost: This is breaking, AFAICT, changing from Capitalized.

metrics.go
klauspost: .. instead maybe here... 
```suggestion
	// TotalTables is the number of tables that had traffic during this window.
	TotalTables int `json:"n"
```

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+	t.BytesIn += other.BytesIn
+	t.BytesOut += other.BytesOut
+}
+
+// IsZero reports whether all counters are zero.
+func (t *TableTrafficMetrics) IsZero() bool {
+	return t.Reads == 0 && t.Writes == 0 && t.BytesIn == 0 && t.BytesOut == 0
+}
+
+// TableMetrics holds windowed traffic for a single table.
```
klauspost: No?

metrics.go
```diff
@@ -3334,6 +3343,102 @@ func (b *BucketAPIMetrics) Merge(other *BucketAPIMetrics) {
...
+type TableMetrics struct {
```
```suggestion
type TopTableMetrics struct {
```

Don't want to "use" the generic name already.

metrics.go
SSushmitha8: It's not top tables at this point. It's all active tables from every node. 
Renamed to MetricsTables

metrics.go
SSushmitha8: We record warehouse, uuid for local traffic and only resolve UUID → name for the final top-5 after ranking, otherwise this we would need to resolve for 1000s of tables(thats's lot of data especially for nested namespace)

metrics.go
SSushmitha8: Tables map[string]TableIOMetrics key is the uuid, we resolve to table names after top 5 is finalized

metrics.go
SSushmitha8: Renamed to TableIOMetrics

metrics.go
SSushmitha8: removed this as its redundant

metrics.go
SSushmitha8: `Tables map[string]TableIOMetrics` is not the final top-5 result  it carries all active tables from each node's local traffic rings before any ranking  happens.  With a slice wouldn't merging be too much time consuming? linear scanning vs hash lookup with map

tier-azure.go
SSushmitha8: Changes from diff branch got added in this

metrics.go
SSushmitha8: TableIOStat      — raw read/write/byte counters for one time window
  TableIOMetrics   — one table's traffic: warehouse + LastMinute/LastHour/LastDay
  TableAPIMetrics  — all active tables keyed by UUID; 
UUID and warehouse name is needed to resolve  the table path later

metrics.go
klauspost: But can there only be one warehouse in a cluster?

klauspost:
Yeah, no, this isn't going to work. I will make a proposal.

klauspost:
Replaced by #618

PR #617, opened 2026-06-04
feat: add D-state threads diag for mc support diag

zveinn:
This PR is part of a 4-repo stack landing the D-state threads diag + the
companion thread / PSI Prometheus metrics. Land in this order so each
downstream PR resolves cleanly:

---

Adds a new diagnostic section that surfaces threads stuck in uninterruptible
disk sleep (D state) through `mc support diag`. Engineers debugging a stuck
cluster can see "1761 threads on iterate_dir" without SSH access.

Schema additions in `health.go` and a new `health-dstate-threads.go`:

- `DStateThreadsDiag`, `DStateThreadGroup`, `DStateThreadInfo` model a
  per-process snapshot grouped by kernel wchan with a small sample of
  stacks per bucket. Per-node identity is supplied by the parent
  `ProcInfo`'s `NodeCommon`.
- New field `ProcInfo.DStateThreads *DStateThreadsDiag`. The pointer
  shape lets servers that don't populate it leave it nil rather than
  emit an empty section.

No wire-format break: existing fields are untouched, only one new
optional JSON key (`dstate_threads`) is added on `ProcInfo`. Older
clients ignore it; newer servers leave it nil when not populated.

### DIFF
--- a/health-dstate-threads.go
+++ b/health-dstate-threads.go
@@ -50,17 +50,17 @@ type DStateThreadGroup struct {
 	Samples    []DStateThreadInfo `json:"samples,omitempty"`
 }
 
-// DStateThreadsDiag is the per-node snapshot of threads in
-// uninterruptible disk sleep, captured by `mc support diag`.
+// DStateThreadsDiag is the per-process snapshot of threads in
+// uninterruptible disk sleep, captured by `mc support diag`. Embedded
+// inside ProcInfo (the process owns its threads); per-node identity
+// comes from the parent ProcInfo's NodeCommon.
 //
-// StuckTotal is the alertable number (Total can be non-zero on a healthy
-// box doing I/O). TotalThreads gives saturation context for dense nodes.
-// WindowSeconds bounds DwellSeconds — a thread reporting
+// StuckTotal is the alertable number (Total can be non-zero on a
+// healthy box doing I/O). TotalThreads gives saturation context for
+// dense nodes. WindowSeconds bounds DwellSeconds — a thread reporting
 // DwellSeconds == WindowSeconds has been in D for at least the entire
 // ring window.
 type DStateThreadsDiag struct {
-	NodeCommon
-
 	Total                 int                 `json:"total"`
 	StuckTotal            int                 `json:"stuck_total,omitempty"`
 	TotalThreads          int                 `json:"total_threads,omitempty"`
--- a/health-dstate-threads.go
+++ b/health-dstate-threads.go
@@ -1,0 +1,50 @@
+//
+// Copyright (c) 2015-2026 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+//
+
+package madmin
+
+// DStateThreadInfo - identifies a single thread caught in uninterruptible
+// disk sleep ("D" state) along with the kernel symbol it was blocked on
+// (wchan) and its kernel stack at the moment of capture.
+type DStateThreadInfo struct {
+	TID   int    `json:"tid"`
+	Name  string `json:"name,omitempty"`
+	Wchan string `json:"wchan,omitempty"`
+	Stack string `json:"stack,omitempty"`
+}
+
+// DStateThreadGroup - groups D-state threads by their shared wchan, with a
+// representative sample of full stacks. Group sizes give an at-a-glance view
+// of where threads are stuck cluster-wide.
+type DStateThreadGroup struct {
+	Wchan   string             `json:"wchan"`
+	Count   int                `json:"count"`
+	Samples []DStateThreadInfo `json:"samples,omitempty"`
+}
+
+// DStateThreadsDiag - per-node snapshot of threads currently in
+// uninterruptible disk sleep, captured as part of `mc support diag`. Helps
+// engineers debugging a stuck cluster see where threads are blocked without
+// requiring SSH access.
+type DStateThreadsDiag struct {
+	NodeCommon
+
+	Total  int                 `json:"total"`
+	Groups []DStateThreadGroup `json:"groups,omitempty"`
+}
--- a/health-dstate-threads.go
+++ b/health-dstate-threads.go
@@ -42,9 +42,18 @@ type DStateThreadGroup struct {
 // uninterruptible disk sleep, captured as part of `mc support diag`. Helps
 // engineers debugging a stuck cluster see where threads are blocked without
 // requiring SSH access.
+//
+// Total is the number of threads in D state at sample time. TotalThreads is
+// the size of the per-thread enumeration the sample was drawn from (i.e. all
+// threads in the MinIO process, regardless of state). The ratio
+// Total/TotalThreads gives consumers a sense of how saturated the process
+// is — a handful of momentarily D-state threads on a dense node with
+// thousands of threads is qualitatively different from the same absolute
+// count on a small node.
 type DStateThreadsDiag struct {
 	NodeCommon
 
-	Total  int                 `json:"total"`
-	Groups []DStateThreadGroup `json:"groups,omitempty"`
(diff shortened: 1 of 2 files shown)


vadmeste (approved):
Good idea!

health.go
```diff
@@ -1481,38 +1482,40 @@ type HealthDataType string
...
+	HealthDataTypeSysMem           HealthDataType = "sysmem"
+	HealthDataTypeSysNet           HealthDataType = "sysnet"
+	HealthDataTypeSysProcess       HealthDataType = "sysprocess"
+	HealthDataTypeSysErrors        HealthDataType = "syserrors"
+	HealthDataTypeSysServices      HealthDataType = "sysservices"
+	HealthDataTypeSysConfig        HealthDataType = "sysconfig"
+	HealthDataTypeSysProductInfo   HealthDataType = "sysproductinfo"
+	HealthDataTypeReplication      HealthDataType = "replication"
+	HealthDataTypeShardsHealth     HealthDataType = "shardshealth"
+	HealthDataTypeSysDStateThreads HealthDataType = "sysdstatethreads"
```
anjalshireesh: Maybe keep it together with the other `sys*` types i.e. immediately after `sysproductinfo`

anjalshireesh (approved):
LGTM, minor nit

health-dstate-threads.go
```diff
@@ -0,0 +1,50 @@
...
+	Wchan   string             `json:"wchan"`
+	Count   int                `json:"count"`
+	Samples []DStateThreadInfo `json:"samples,omitempty"`
+}
+
+// DStateThreadsDiag - per-node snapshot of threads currently in
+// uninterruptible disk sleep, captured as part of `mc support diag`. Helps
+// engineers debugging a stuck cluster see where threads are blocked without
+// requiring SSH access.
+type DStateThreadsDiag struct {
```
vadmeste, line 63: I think it is a good idea to add the total number of threads as well..  Sometimes a customer put a lot of drives in a single machine and I think that in practice, it is possible to see few threads in D state because threads naturally go to D state but just for very short period but can still be visibile

vadmeste (approved):

vadmeste (approved):

anjalshireesh (approved):

zveinn:
@vadmeste I added some metrics to keep track of task id's (thread executions) over time .. specifically the ones that dwell too long in D state
```
$ ./mc admin prometheus metrics m0 debug --api-version v3 | grep dwell
# HELP minio_debug_system_threads_dstate_count D-state threads with continuous-on-same-wchan dwell at least dwell_ge seconds
minio_debug_system_threads_dstate_count{dwell_ge="0",server="127.0.0.1:9000"} 0
minio_debug_system_threads_dstate_count{dwell_ge="10",server="127.0.0.1:9000"} 0
minio_debug_system_threads_dstate_count{dwell_ge="15",server="127.0.0.1:9000"} 0
minio_debug_system_threads_dstate_count{dwell_ge="30",server="127.0.0.1:9000"} 0
minio_debug_system_threads_dstate_count{dwell_ge="5",server="127.0.0.1:9000"} 0
minio_debug_system_threads_dstate_count{dwell_ge="60",server="127.0.0.1:9000"} 0
# HELP minio_debug_system_threads_dstate_window_seconds Look-back window of the D-state dwell ring buffer in seconds
```

klauspost:
@zveinn This should be in persisted [process metrics](https://github.com/minio/madmin-go/blob/4acdae25134aa1373fdcb68415a54494c11aeb08/metrics.go#L2858) as well. 

Would you like to add it, or should I pick it up afterwards?

health.go
```diff
@@ -1276,17 +1276,18 @@ func (p *ProcInfo) AddProcInfo(metrics *ProcessMetrics) {
...
+	Partitions     []Partitions        `json:"partitions,omitempty"`
+	OSInfo         []OSInfo            `json:"osinfo,omitempty"`
+	MemInfo        []MemInfo           `json:"meminfo,omitempty"`
+	ProcInfo       []ProcInfo          `json:"procinfo,omitempty"`
+	NetInfo        []NetInfo           `json:"netinfo,omitempty"`
+	SysErrs        []SysErrors         `json:"errors,omitempty"`
+	SysServices    []SysServices       `json:"services,omitempty"`
+	SysConfig      []SysConfig         `json:"config,omitempty"`
+	ProductInfo    []ProductInfo       `json:"productinfo,omitempty"`
+	DStateThreads  []DStateThreadsDiag `json:"dstate_threads,omitempty"`
```
anjalshireesh: Does it make sense to have this (`DStateThreads`) under `ProcInfo` ? cc @klauspost @harshavardhana

zveinn:
> @zveinn This should be in persisted [process metrics](https://github.com/minio/madmin-go/blob/4acdae25134aa1373fdcb68415a54494c11aeb08/metrics.go#L2858) as well.
> 
> Would you like to add it, or should I pick it up afterwards?

yea, let's get this into diag and prometheus and then into api metrics. 100%

health.go
klauspost: Yes, that is what I eluded to above. We would need a mergeable version, though.

anjalshireesh: @zveinn can you please move `DStateThreads` to inside `ProcInfo` ?

health.go
```diff
@@ -1495,40 +1496,42 @@ type HealthDataType string
...
+	HealthDataTypeSysDriveHw       HealthDataType = "sysdrivehw"
+	HealthDataTypeSysOsInfo        HealthDataType = "sysosinfo"
+	HealthDataTypeSysMem           HealthDataType = "sysmem"
+	HealthDataTypeSysNet           HealthDataType = "sysnet"
+	HealthDataTypeSysProcess       HealthDataType = "sysprocess"
+	HealthDataTypeSysErrors        HealthDataType = "syserrors"
+	HealthDataTypeSysServices      HealthDataType = "sysservices"
+	HealthDataTypeSysConfig        HealthDataType = "sysconfig"
+	HealthDataTypeSysProductInfo   HealthDataType = "sysproductinfo"
+	HealthDataTypeSysDStateThreads HealthDataType = "sysdstatethreads"
```
anjalshireesh: Won't be required if this becomes part of `sysprocess` info

zveinn: I don't think we want this.. ProcInfo is an on-demand scraping and ProcInfo is not inside prometheus. We can add it there later as an addition, but we want/need these to be scrapable on an interval and available inside prometheus.

capacity-forecast.go
```diff
@@ -47,9 +47,8 @@ type CapacityForecast struct {
 	// GrowthBytesPerDay is the Kalman filter slope expressed in bytes per
 	// day, projected from the daily snapshots in the circular buffer.
 	// It is independent of DailySnapshotCount: the filter is recency
-	// weighted, not a delta between two endpoints. It is a float64 so a
-	// slow, sub-byte-per-day slope is not truncated to zero.
-	GrowthBytesPerDay float64 `json:"growthBytesPerDay"`
+	// weighted, not a delta between two endpoints.
+	GrowthBytesPerDay int64 `json:"growthBytesPerDay"`
```
anjalshireesh: This seems unrelated to d-state threads diag. Did it come in by mistake?

health.go
zveinn: chekc my comment on the other mention of ProcInfo, it's not being scraper nor is it inside prometheus. We can add the DStateThreadsDiag as a sub structs in proc info later when we add this into real time metrics.

capacity-forecast.go
zveinn: Yes, this is something that I was about to clean out .. it's because of a missing addition in capacity, shouldn't be here at all

health.go
anjalshireesh: AFAIK these types are exclusively used by health diagnostics. If the D-state threads info is related to the process, then there is no need to add it separately as a new type and it can become part of the "process info" in the diagnostics data. Is there a strong reason to keep this out of process info?

health.go
anjalshireesh: In that case it might be better to do both together. I think we should not add a new category of diagnostics data especially when it fits into an existing category (`sysprocinfo`). In the diag data structure also it would be better to keep this in its logical place (inside `procinfo`)

health.go
harshavardhana: @zveinn

health.go
zveinn: We can't keep this inside the ProcInfo structure only since we NEED this for prometheus, ProcInfo is no in prometheus and we're not sending that info out using prometheus. We can also use the Dstate structure as a field inside ProcInfo later for the realtime metrics.

Regarding the category, i'll move the thread info into system/cpu

health.go
zveinn: We will need to have dstate threads seperately for prometheus since ProcInfo isin't a part of prometheus, but it can live inside ProcInfo as an embeded struct for the "mc support diag" path, will make those changes today.

anjalshireesh (approved):
LGTM

klauspost (approved):

PR #638, opened 2026-07-16
feat: add MakeCortexBktOp for gated cortex-bucket site replication

harshavardhana:
## Description

Adds a `MakeCortexBktOp` peer bucket-create operation for AIStor Memory
cortex buckets, mirroring `MakeWarehouseBktOp` (#637).

A Memory **cortex** is always also a Tables **warehouse** (a super-bucket),
so this single op provisions both capabilities on the peer. Encoding the
bucket type as the operation — rather than an ad-hoc `isCortex` form field —
matches the pattern established for warehouses in #637.

## Changes

- `MakeCortexBktOp BktOp = "make-cortex"` constant.
- `SRPeerBucketOps` forwards bucket options (`versioningEnabled`,
  `forceCreate`, `createdAt`, …) for the new op.

## Compatibility

Like `MakeWarehouseBktOp`, a peer that predates cortex replication has no
route for this op and rejects it, so the cortex is withheld until that peer
is upgraded — no partial/broken state on old peers.

## Consumer

Used by MinIO AIStor's Memory API (super-bucket site replication); the
server side switches its SR make-bucket op to `MakeCortexBktOp` once this is
released, replacing the interim `isCortex` opts field.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating Cortex buckets through cluster bucket operations.
  * Cortex creation now includes the related tables warehouse capability.
  * Bucket operation options are now forwarded correctly when creating Cortex buckets.
* **Tests**
  * Added coverage to verify Cortex bucket operation query parameters and options forwarding.

### DIFF
--- a/cluster-commands.go
+++ b/cluster-commands.go
@@ -242,6 +242,12 @@ const (
 	// predates warehouse replication has no route for this op and rejects it, so
 	// the warehouse is withheld until that peer is upgraded
 	MakeWarehouseBktOp BktOp = "make-warehouse"
+	// make an AIStor Memory cortex bucket and enable versioning. A cortex is
+	// always also a Tables warehouse (a super-bucket), so this op implies the
+	// warehouse capability too. A peer that predates cortex replication has no
+	// route for this op and rejects it, so the cortex is withheld until that peer
+	// is upgraded.
+	MakeCortexBktOp BktOp = "make-cortex"
 	// add replication configuration
 	ConfigureReplBktOp BktOp = "configure-replication"
 	// delete bucket (forceDelete = off)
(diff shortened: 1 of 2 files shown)


harshavardhana:
## Fixes Applied Successfully

Added a regression test based on 1 CodeRabbit feedback item.

**Files modified:**
- `cluster-commands_test.go`

**Commit:** `680bd7a`

`TestSRPeerBucketOpsForwardsCortexOpts` asserts a `MakeCortexBktOp` peer
request forwards versioning / force-create / created-at options. `go test`
passes.

harshavardhana:
## Fixes Applied Successfully

Addressed 1 CodeRabbit feedback item.

**Files modified:**
- `cluster-commands_test.go`

**Commit:** `d5a9d4c`

Capture the request query via a buffered channel instead of a shared
variable, removing the cross-goroutine data-race risk. `go test -race` passes.

PR #639, opened 2026-07-22
tables: run snapshot-expiry endpoint

twuebi:
adds an endpoint which synchronously runs snapshot expiration for a given table

## Summary by CodeRabbit

- **New Features**
  - Added a new admin API to manually trigger snapshot expiration for a specific table via an HTTP POST.
  - Supports optional overrides for maximum snapshot age and minimum snapshots to keep (only provided values are sent).
  - Returns the number of snapshots expired as the operation result.
- **Tests**
  - Added coverage for request formation (method/path/query/body), response decoding, omission of unset overrides, and proper error handling for non-2xx responses.

### DIFF
--- a/tables-snapshot-expiry.go
+++ b/tables-snapshot-expiry.go
@@ -1,0 +1,85 @@
+//
+// Copyright (c) 2015-2026 MinIO, Inc.
+//
+// This file is part of MinIO Object Storage stack
+//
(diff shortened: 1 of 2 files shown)


rraulinio (commented):
Maybe add some unit tests please?

rraulinio (approved):

PR #642, opened 2026-08-02
feat: add MemPerf() for memory bandwidth validation

harshavardhana:
Adds `MemPerf()` plus its result types, calling a new `/speedtest/mem` admin endpoint. Server side follows in a separate PR.

Memory bandwidth is what caps large-object throughput on many-core storage nodes, and it is the one resource no utilization counter shows — DMA from a drive or NIC never touches a core, so CPU, disk and network can all read idle while DRAM is saturated.

The result carries the measurement next to the topology it should be judged against, so `Efficiency()` distinguishes a genuinely saturated node from a mispopulated or downclocked one. Theoretical bandwidth is derived from the *configured* clock rather than the rated one — 5600 MT/s parts pinned at 4800 are common, and using the rated speed would make a healthy node look broken.

Topology capture is best effort and records its own provenance (`dmi` / `edac` / `unavailable`), so an unreadable source is never mistaken for a measured zero.

## Summary by CodeRabbit

* **New Features**
  * Added a memory-bandwidth performance test through the administration interface.
  * Supports configurable test duration, thread count, and buffer size.
  * Reports performance results, memory topology, DIMM details, efficiency, and downclocking status.
  * Provides clear handling for unavailable topology information, invalid settings, cancelled requests, and service errors.
* **Tests**
  * Added coverage for request options, response handling, error scenarios, efficiency calculations, and downclock detection.

### DIFF
--- a/perf-mem.go
+++ b/perf-mem.go
@@ -88,6 +88,11 @@ type MemPerfNodeResult struct {
 // the topology could not be read. Values below roughly 0.65 are worth
 // investigating before trusting any other benchmark from the node.
 func (r MemPerfNodeResult) Efficiency() float64 {
+	// A theoretical figure with no source behind it is not a measurement;
+	// reporting a ratio from it would dress up a guess as data.
+	if r.TopologySource == "" || r.TopologySource == MemPerfTopologyUnavailable {
+		return 0
+	}
 	if r.TheoreticalBps == 0 {
 		return 0
 	}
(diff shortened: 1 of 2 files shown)


harshavardhana:
## Fixes Applied Successfully

Fixed 2 file(s) based on 3 CodeRabbit feedback item(s).

**Files modified:**
- `perf-mem.go`
- `perf-mem_test.go`

**Commit:** `67e7257`

- `Efficiency()` now returns 0 unless a topology source backs the theoretical figure — used a deny-list on unavailable/unset rather than an allow-list, so a future legitimate source is not silently zeroed.
- `MemPerf()` rejects negative `Duration`/`Threads` via `ErrInvalidArgument` instead of falling through to the server default (which for threads is every core). `BufferSize` is unsigned and needs no check.
- Added `perf-mem_test.go`: request shape, option serialization and zero-omission, negative-option rejection without issuing a request, multi-node decode, non-200, malformed JSON, cancelled context, plus `Efficiency()` and `Downclocked()`.

The latest autofix changes are on the `feat/mem-perf` branch.

harshavardhana:
## Fixes Applied Successfully

Fixed 1 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `perf-mem_test.go`

**Commit:** `9d1fbd3`

- **Assert the invalid-argument error contract** — `TestMemPerfNegativeOpts` asserted only that some error occurred, which a dial failure would satisfy just as well as validation. It now asserts an `ErrorResponse` carrying `InvalidArgument`.
- **Add focused tests for `AdminClient.MemPerf`** — no action needed; this is the same thread re-anchored to a new line after the previous commit. `perf-mem_test.go` already covers query serialization, zero-option omission, invalid inputs, non-200, malformed JSON and context cancellation.

The latest autofix changes are on the `feat/mem-perf` branch.

PR #651, opened 2026-08-27
Fix node merges bloating node count

klauspost:
When merging within a node the node count should be preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected API and replication metrics so node counts are no longer inflated when combining time periods or targets.
  * Preserved additive totals for request and event counters while reporting accurate node counts.
  * Updated last-day, last-hour, and overall metric views to display correct counts across targets and time segments.
  * Improved handling of empty, missing, partially reported, and overlapping metric data.

### DIFF
--- a/metrics.go
+++ b/metrics.go
@@ -2478,30 +2478,36 @@ func (m *ReplicationMetrics) Merge(other *ReplicationMetrics) {
 }
 
 // AllTargets returns aggregated stats for all targets.
+//
+// Node counts cannot be folded exactly along the target axis. A node gains a
+// map entry for a target only once it has processed an event for that target,
+// so reporter sets differ per target and the sums Merge leaves behind bound the
+// union of reporters without pinning it: max over targets is a lower bound, and
+// the number of nodes that responded is an upper one. Two targets reported by
+// one distinct node each give every target Nodes == 1 while two nodes really
+// contributed, so the lower bound undercounts; this clamps the sum to the upper
+// bound instead, which is exact whenever every responding node reported at
+// least one target, and never reads below the per-target maximum.
 func (m *ReplicationMetrics) AllTargets() ReplicationTargetStats {
 	var dst ReplicationTargetStats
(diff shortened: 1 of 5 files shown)


0xMALVEE (approved):

metrics.go
```diff
@@ -2461,9 +2480,28 @@ func (m *ReplicationMetrics) Merge(other *ReplicationMetrics) {
...
+		nodes = max(nodes, v.Nodes)
+		lastMinute = max(lastMinute, v.LastMinute.Nodes)
+		lastHour = max(lastHour, v.LastHour.Nodes)
+		sinceStart = max(sinceStart, v.SinceStart.Nodes)
+		if v.LastDay != nil {
+			days = append(days, v.LastDay)
+		}
+	}
+	// Merge sums Nodes, which is only correct along the node axis. The same
+	// nodes report every target, so folding across targets must not sum;
```
rraulinio: **Preserve the union of reporting nodes across targets**

This assumes that every target is reported by the same set of nodes, but AIStor builds each node's target map lazily when that node processes an event for that target. The reporter sets can therefore differ.

For example, if node 1 reports only target A and node 2 reports only target B, both targets have `Nodes == 1`, while two distinct nodes contributed to the all-target aggregate. Taking the maximum returns `1` instead of `2`. The same undercount affects `LastMinute`, `LastHour`, `SinceStart`, and aligned last-day segments.

The current per-target counts do not contain enough information to determine overlaps, thus should compute/preserve the reporter union before node identity is discarded, or carry an aggregation-safe count from the producer. Should also add a regression test with disjoint or partially overlapping reporter sets; the current tests only cover identical sets.

rraulinio (changes requested):

metrics.go
klauspost: Clamping to nodes. This may not be 100% exact, but should be closer than now.

rraulinio (approved):

PR #656, opened 2026-09-08
nmav: Improve process/cpu/mem stats

klauspost:
Ensure there are per node averages for process/cpu/mem stats.

Fix various minor inconsistencies in request averages.

## Summary by CodeRabbit

* **New Features**
  * Added last-hour and last-day navigation for runtime and memory metrics.
  * Added per-segment and `_ALL` aggregate views for process, memory, and runtime statistics.
  * Runtime displays now include uptime metrics and clearer per-node summaries.

* **Bug Fixes**
  * Improved API, replication, process, and runtime rate calculations using actual reporting windows.
  * Improved per-node averages, percentage shares, cumulative counter handling, and histogram formatting.
  * Corrected time-segment labels, navigation consistency, and duration and memory formatting.

### DIFF
--- a/mnav/mem.go
+++ b/mnav/mem.go
@@ -117,7 +117,7 @@ func (r *memRows) total(label string, v, whole uint64) {
 	if v == 0 {
 		return
 	}
-	r.add(label, fmtBytes(float64(v))+qualify(r.nodes, float64(v), float64(whole), fmtBytes))
+	r.add(label, fmtBytes(float64(v))+qualify(r.nodes, "node", float64(v), float64(whole), fmtBytes))
 }
 
 func (node *MemMetricsNavigator) GetLeafData() map[string]string {
@@ -333,7 +333,7 @@ func (node *MemSwapNode) GetLeafData() map[string]string {
 	// Rendered even at zero: "none in use" is the reassurance being looked for,
 	// and hiding it makes a healthy cluster look unreported.
 	used := info.SwapSpaceTotal - info.SwapSpaceFree
-	r.add("Used", fmtBytes(float64(used))+qualify(r.nodes, float64(used), float64(info.SwapSpaceTotal), fmtBytes))
+	r.add("Used", fmtBytes(float64(used))+qualify(r.nodes, "node", float64(used), float64(info.SwapSpaceTotal), fmtBytes))
 	r.total("Free", info.SwapSpaceFree, info.SwapSpaceTotal)
 	if info.Total > 0 {
 		r.add("Swap : RAM", fmt.Sprintf("%.2f : 1", float64(info.SwapSpaceTotal)/float64(info.Total)))
--- a/mnav/mem.go
+++ b/mnav/mem.go
@@ -444,8 +444,10 @@ var memWindowRows = []memWindowRow{
 // Every field is summed over the samples folded in -- N of them, one per node per
 // segment -- so the quotient is a mean per sample and one sample covers one
 // interval. That makes interval the divisor for every rate here even when the
-// whole window is rendered; segments is what turns N back into a node count and
-// scales a counter's total to one node's share of the window.
+// whole window is rendered; segments is how many slots carried a sample -- not
+// how many the window holds, since nothing accrued in the empty ones -- and is
+// what turns N back into a node count and scales a counter's total to one node's
+// share.
 func memSegmentRows(seg madmin.MemSegment, interval, segments int, coverage string) map[string]string {
 	if seg.N == 0 {
 		return map[string]string{"Status": "no node reported this time segment"}
@@ -545,6 +547,19 @@ func (node *MemLastDayNode) hasSegments() bool {
 	return node.segmented != nil && node.segmented.Interval > 0 && len(node.segmented.Segments) > 0
 }
 
+// reportedSegments counts the slots that carry a sample. A window's slots are not
+// all populated -- a restart, a startup or a node joining late leaves gaps -- and
+// nothing accrued in those.
+func (node *MemLastDayNode) reportedSegments() int {
+	var n int
+	for i := range node.segmented.Segments {
+		if node.segmented.Segments[i].N > 0 {
+			n++
+		}
+	}
+	return n
+}
+
 func (node *MemLastDayNode) wholeSecs() int {
 	return node.segmented.Interval * len(node.segmented.Segments)
 }
@@ -587,7 +602,7 @@ func (node *MemLastDayNode) GetChild(name string) (MetricNode, error) {
 		return &MemTimeSegmentNode{
 			segment:  node.segmented.Total(),
 			interval: node.segmented.Interval,
-			segments: len(node.segmented.Segments),
+			segments: node.reportedSegments(),
 			coverage: windowCoverage(node.segmented.FirstTime, node.wholeSecs()),
 			flags:    node.flags, parent: node, path: node.path + "/" + name,
 		}, nil
@@ -613,7 +628,7 @@ func (node *MemLastDayNode) GetLeafData() map[string]string {
 	// The aggregate only: each segment states its own figures in the description
 	// of the child that opens it.
 	return memSegmentRows(node.segmented.Total(), node.segmented.Interval,
-		len(node.segmented.Segments), windowCoverage(node.segmented.FirstTime, node.wholeSecs()))
+		node.reportedSegments(), windowCoverage(node.segmented.FirstTime, node.wholeSecs()))
 }
 
 // MemTimeSegmentNode is one segment of a memory window, or the whole window.
--- a/mnav/mem.go
+++ b/mnav/mem.go
@@ -20,7 +20,7 @@ package mnav
 import (
 	"fmt"
 	"strconv"
-	"time"
+	"strings"
 
 	"github.com/dustin/go-humanize"
 	"github.com/minio/madmin-go/v4"
@@ -66,7 +66,8 @@ func (node *MemMetricsNavigator) GetChildren() []MetricChild {
 		MetricChild{Name: "system", Description: "System memory details (cache, buffers, shared)"},
 		MetricChild{Name: "swap", Description: "Swap space information and utilization"},
 		MetricChild{Name: "limits", Description: "Memory limits and cgroup configuration"},
-		MetricChild{Name: "last_day", Description: "Last 24h memory statistics"},
+		MetricChild{Name: "last_hour", Description: "Memory levels and kernel reclaim over the last hour, by time segment"},
+		MetricChild{Name: "last_day", Description: "Memory levels and kernel reclaim over the last day, by time segment"},
 	)
 	if node.mem.VMStat != nil {
 		children = append(children,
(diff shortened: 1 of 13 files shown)


0xMALVEE (approved):

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
+				s.max = cur
+			}
+			s.sum += cur
+			s.samples++
+			if row.counter {
+				// The span is measured from the segment that last reported,
+				// not from the one before this in the slice: a gap of three
+				// quarter hours means the increase took forty-five minutes.
+				if before, seen := prev[row.key]; seen && cur >= before.value {
+					d := segDelta{cur - before.value, (i - before.index) * seg.Interval}
```
rraulinio: **Do not difference runtime counters that are already bucket deltas**

Historical runtime producer stores `current - previous` in each `RuntimeSegment` for GC cycles and pause counts. Subtracting adjacent segments again therefore calculates the change between two bucket deltas, not activity during the bucket.

For example, three buckets containing 10 GC cycles each currently produce zero activity because every adjacent difference is zero; the correct window total is 30. A sequence such as `[10, 12, 8]` reports only `+2` and discards the final 8 as a reset.

Should treat historical runtime counters as per-segment deltas, sum them directly, derive each rate from the segment interval, and update the tests to use producer-shaped values rather than cumulative `[1000, 1010, ...]` fixtures.

mnav/process.go
```diff
@@ -854,142 +786,172 @@ func (node *ProcessLastDayNode) GetChild(name string) (MetricNode, error) {
...
+		r.add("Uninterruptible", fmtCount(float64(seg.ThreadsD)/n)+" threads per process")
+	}
+
+	// I/O and the fault and switch counters accrue inside the window, so they
+	// carry a rate rather than a level.
+	ioRow := func(label string, v float64, render func(float64) string, unit string) {
+		if v <= 0 {
+			return
+		}
+		perProc := v / n
```
rraulinio, line 900: **Preserve absolute counter totals in the `_ALL` process view**

For `_ALL`, `seg` is the sum of every segment while `N` is also summed as `processes * segments`. Consequently, `v / n` is the average for one process during one segment, not the per-process total over the complete window.

For example, fourteen segments containing 900 MB read per process currently render `_ALL` as `900 MB per process`, although the combined window contains 12.6 GB per process. CPU times, operation counts, faults, and context switches have the same issue. The rate calculation is correct because the per-segment average is divided by one interval, but the absolute value is mislabeled.

Should calculate the absolute value using the number of reported segments, or explicitly label it as a per-segment average. The test currently requiring `_ALL` to equal one segment should also distinguish gauges from delta counters.

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
+// clock since the process started, so dividing it by GOMAXPROCS gives that wall
+// clock back -- both summed over the same nodes, so the quotient needs no node
+// count. The derivation breaks where GOMAXPROCS changed mid-flight, which Go 1.25
+// does on its own when a container's CPU limit moves; the reported value does not.
+//
+// Either way it is a mean: a node restarted ten minutes ago pulls the figure down
+// instead of showing up on its own, which is why every rate taken from it is
+// labelled per node rather than presented as a fleet fact.
+func (v goVals) uptime() (d time.Duration, reported, ok bool) {
+	if v.upNodes > 0 && v.up > 0 {
```
rraulinio: **Do not rate all-node counters against uptime from only a subset of nodes**

During a mixed-version rollout, `UintMetrics` and `FloatMetrics` still contain values from all `N` nodes, while `UptimeSecs / UptimeNodes` covers only nodes supporting the new uptime fields. Using that partial uptime for all-node counters can inflate rates dramatically.

For example, if an old node has 86,400 cycles over 24 hours and a newly upgraded node has 60 cycles over 60 seconds, but only the new node reports uptime, the current calculation reports roughly 720 cycles/s instead of 1 cycle/s.

Should use reported uptime as the counter denominator only when `UptimeNodes == N`. Otherwise, use the existing all-node derivation or omit the rate until the populations align. The partial-rollout test should validate a counter rate as well as the displayed uptime.

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
-		for i, count := range gcPauseHist.Counts {
-			if i < len(gcPauseHist.Buckets)-1 && count > 0 {
-				bucketMidpoint := (gcPauseHist.Buckets[i] + gcPauseHist.Buckets[i+1]) / 2
-				totalSum += bucketMidpoint * float64(count)
-			}
+		count += c
+		lo, hi := h.Buckets[i], h.Buckets[i+1]
+		// The outermost buckets are unbounded; the finite edge is the only
+		// estimate either of them offers.
+		if math.IsInf(lo, -1) {
```
rraulinio: **Handle the finite sentinels used for unbounded histogram buckets**

The producer replaces histogram `-Inf` and `+Inf` boundaries with `-math.MaxFloat64` and `math.MaxFloat64` before putting them on the wire. Therefore these `math.IsInf` checks never recognize the actual values received by madmin-go.

When an outer bucket contains samples, its midpoint includes `MaxFloat64`, causing the calculated average to overflow `time.Duration`; an upper percentile is also presented as bounded instead of as a lower bound.

Should recognize the producer's finite sentinels as unbounded boundaries, or preserve that information explicitly in the wire schema, and add a producer-shaped outer-bucket test.

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
-func (node *GCMetricsNode) ShouldPauseRefresh() bool {
-	return false
-}
-
-func (node *GCMetricsNode) GetChild(_ string) (MetricNode, error) {
-	return nil, fmt.Errorf("GC metrics node has no children")
+// histQuantile returns the bucket edge the qth sample falls in. bounded is false
+// in the open-ended top bucket, where the edge is a lower bound.
+func histQuantile(h metrics.Float64Histogram, count uint64, q float64) (edge float64, bounded, ok bool) {
+	want := uint64(float64(count) * q)
```
rraulinio: **Use a ceiling rank when selecting histogram percentiles**

Converting `count * q` directly to `uint64` floors the desired rank. With two samples in different buckets, p99 becomes rank 1 and incorrectly selects the lower bucket instead of the second sample.

Should use a ceiling-based rank, clamped to `[1, count]`, and add small-sample tests such as two observations in separate buckets.

rraulinio (changes requested):

shtripat (commented):
(placeholder while anchoring comment positions -- see the full review below.)

shtripat (commented):
(placeholder while anchoring comment positions -- see the full review below.)

shtripat (commented):
(placeholder while anchoring comment positions -- see the full review below.)

shtripat (commented):
(placeholder while anchoring comment positions -- see the full review below.)

mnav/runtime.go
klauspost: ack

mnav/process.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/process.go
```diff
@@ -296,61 +343,40 @@ func (node *ProcessCPUTimesNode) GetLeafData() map[string]string {
...
-				node.cpuTimes.Guest, (node.cpuTimes.Guest/totalTime)*100)
-		}
-		if node.cpuTimes.GuestNice > 0 {
-			data["Guest Nice Time"] = fmt.Sprintf("%.2f seconds (%.1f%%)",
-				node.cpuTimes.GuestNice, (node.cpuTimes.GuestNice/totalTime)*100)
-		}
-	}
-
-	return data
+	c := node.cpuTimes
```
shtripat, line 346: None of the five process subsection leaves get exercised by a test: `ProcessCPUTimesNode.GetLeafData` (here), `ProcessMemoryInfoNode.GetLeafData` (mnav/process.go:414), `ProcessCtxSwitchesNode.GetLeafData` (mnav/process.go:534), `ProcessPageFaultsNode.GetLeafData` (mnav/process.go:587), and `ProcessMemoryMapsNode.GetLeafData` (mnav/process.go:642) all show 0% coverage. process_test.go only navigates to `process`, `process/io`, and `process/last_day` -- never `process/cpu`, `process/memory`, `process/context_switches`, `process/page_faults`, or `process/mem_maps`. That's the CPU busy-percent math, the mapped-memory share breakdown, and the voluntary/involuntary split all shipping untested. Worth adding one test per node the way `TestProcessSubsectionRatesUseParentUptime` already does for `process/io`.

mnav/mem.go
```diff
@@ -211,61 +221,21 @@ func (node *MemUsageNode) GetLeafData() map[string]string {
...
-		} else if utilizationRatio > 0.6 {
-			efficiencyNote = "Good utilization"
-		} else {
-			efficiencyNote = "Low utilization"
-		}
-		data["Efficiency"] = efficiencyNote
-	}
-
-	return data
+	if info.Total == 0 {
```
shtripat, line 225: Same gap on the memory side: `MemUsageNode.GetLeafData` (here), `MemSystemNode.GetLeafData` (mnav/mem.go:273), `MemSwapNode.GetLeafData` (mnav/mem.go:323), and `MemLimitsNode.GetLeafData` (mnav/mem.go:376) are all 0% covered -- nothing navigates to `mem/usage`, `mem/system`, `mem/swap`, or `mem/limits`. The rewritten headroom/limit logic in `MemLimitsNode` and the reclaimable-cache row in `MemSystemNode` go in with no test backing them.

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
+	}
+	// Only the node count made it, so this Go version reports nothing this
+	// section knows how to render.
+	if r.n == 1 {
+		return map[string]string{"Status": "no " + node.kind + " metrics in this sample"}
+	}
+	return r.data
+}
+
+func (node *runtimeSectionNode) gcRows(v goVals, r *runtimeRows) {
```
shtripat, line 495: `gcRows` (here), `schedulerRows` (mnav/runtime.go:533), and `syncRows` (mnav/runtime.go:595) are all 0% covered -- runtime_test.go exercises `go/memory` and `go/cpu_classes` but never `go/gc`, `go/scheduler`, or `go/sync`. That's the finalizer/cleanup queue tracking, the per-thread goroutine ratio, and the blocked-rate calculation in `syncRows` all shipping untested.

mnav/runtime.go
shtripat: `want` truncates instead of rounding up, so p99 can collapse onto the same bucket as p50 whenever the rank falls right at a boundary. Checked it directly: 2 samples split one-each across buckets `[0,10]` and `[10,20]`, and `histQuantile(h, 2, 0.99)` returns edge `10` -- the same as `histQuantile(h, 2, 0.5)` -- when it should land in the second bucket. Nearest-rank percentile estimation wants `rank = ceil(q*N)`, not `floor(q*N)`, and `uint64(count*q)` truncates. Only shows up at small N, which is exactly what a quiet 15-minute segment's GC-pause histogram tends to have -- `TestRuntimeHistogramKeepsNanoseconds` only checks N=672 so it doesn't catch this.
```suggestion
	want := uint64(math.Ceil(float64(count) * q))
```

mnav/rpc.go
```diff
@@ -710,72 +710,6 @@ func (node *RPCLastDayHandlerNode) GetChild(name string) (MetricNode, error) {
...
-			rpc:    node.rpc,
-			parent: node,
-			path:   node.path + "/" + name,
-		}, nil
-	default:
-		return nil, fmt.Errorf("connection child not found: %s", name)
-	}
-}
-
 // RPCConnectionSummaryNode shows connection summary details
```
shtripat: `RPCConnectionSummaryNode` has no constructor left anywhere in the tree -- its only caller, `RPCConnectionsNode.GetChild("summary")`, was deleted in this same diff. `grep -rn "RPCConnectionSummaryNode"` turns up nothing but the type's own method definitions (through line 763). Worth deleting it along with the rest of the RPC cleanup instead of leaving it dead.

shtripat (commented):
Went through this in detail: build, full test suite, and race tests are all clean, and package coverage went up rather than down. Left inline notes on a few things -- the process/mem/runtime detail subsections got substantially rewritten but several of them have zero test coverage, there's a real off-by-one in histQuantile's percentile rank calc that shows up at small sample counts, and one type is dead code left over from the RPC cleanup.

mnav/runtime.go
```diff
@@ -19,614 +19,926 @@ package mnav
...
-				bucketMidpoint := (schedLatency.Buckets[i] + schedLatency.Buckets[i+1]) / 2
-				totalSum += bucketMidpoint * float64(count)
+		v := segmentVals(seg.Segments[i])
+		w.deltas[i] = make(map[string]segDelta, len(runtimeWindowRows))
+		for _, row := range runtimeWindowRows {
+			cur, ok := v.node(row.key)
+			if !ok {
+				continue
 			}
+			s := w.stats[row.key]
```
shtripat, line 696: nit: `s.min`/`s.max`/`s.sum`/`s.samples` get updated here for counter rows too, but `summarize()` (line 737) returns early via `s.delta` for counters and never reads them. Harmless, just wasted work every time a window renders.

klauspost: Yeah, noone will notice.

klauspost:
@shtripat Please adjust your LLM to not post "placeholders".

shtripat:
> @shtripat Please adjust your LLM to not post "placeholders".

yes, first time saw this glitch. Will do needful. thanks

mnav/process.go
klauspost: ack

mnav/mem.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/rpc.go
klauspost: ack

mnav/runtime.go
klauspost: ack

mnav/process.go
```diff
@@ -854,142 +786,176 @@ func (node *ProcessLastDayNode) GetChild(name string) (MetricNode, error) {
...
+		if v <= 0 {
+			return
+		}
+		// v/n is one process's figure for one segment; the window's is that
+		// times the segments folded in. The share is the same either way, since
+		// both sides scale together.
+		perProc := v / n
+		if interval > 0 {
+			r.add(label, fmt.Sprintf("%s (%s of one core)",
+				fmtSecs(perProc*float64(segments)), fmtPct(perProc, float64(interval))))
```
rraulinio, line 860: **Do not extrapolate counters across unreported time segments**

`segments` is the total number of slots in the window, while `seg.N` and the counter values only include populated slots. Multiplying `v / N` by every slot therefore invents activity for slots where `N == 0`.

For example, a two-slot window containing one 900 MB read sample and one empty slot currently renders `_ALL` as 1.8 GB per process, although only 900 MB was observed. Empty slots can occur during startup, restarts, or partial node participation. The same calculation is present in `memSegmentRows` at `mnav/mem.go:470`.

Should aggregate counters per reported segment (`value / N`) and sum those results, rather than multiplying the combined average by the total window length. A regression test with an `N == 0` gap would cover this.

rraulinio (changes requested):

mnav/process.go
klauspost: ack

rraulinio (approved):

shtripat (approved):

PR #658, opened 2026-09-15
feat: add the AIStor Files export status query

HJLebbink:
## Description

Adds the client type and method for `GET /minio/admin/v4/query/files-exports`,
which reports the status of named AIStor Files gateway exports on every node a
cluster can reach.

The endpoint landed in miniohq/aistor#7206 with its reply type defined in that
repository and encoded as JSON, so no client outside the server could read it.
This change moves the type here and puts it on MessagePack, matching the other
v4 query endpoints. `mc` is the consumer that needs it.

The server-side switch to this type is a separate change in aistor and depends
on a tag of this one.

## API

`AdminClient.FilesExportsQuery(ctx, exportIDs []uint64)` returns
`FilesExportsQueryResponse`:

- `FilesExportsQueryResponse` holds `Results`, `Count`, `Total`, `Unreachable`
  and `PeersNotQueried`. The endpoint does not paginate, so `Count` and `Total`
  differ only by the nodes that could not be asked.
- `FilesNodeStatus` is one node's answer, carrying `Reach`, `Detail`,
  `SocketPath`, `Truncated` and its per-export results.
- `FilesExportResult` is one export's outcome. At most one of `Status`,
  `NotHeld` and `Error` carries it.
- `FilesExportStatus` is the lease and capacity document the daemon reported.
- `FilesNodeReach` distinguishes five outcomes per node: the node answered, the
  node runs no gateway daemon, the gateway refused the server process, the
  gateway rejected the request itself, and nothing was established. A node
  running no gateway is an ordinary state, while a socket that will not admit
  the server is a deployment fault, so they are separate values.

Two constants are exported because callers need them to interpret a reply:
`MaxFilesExportIDsPerQuery` is the 512-id bound the server also applies, and
`FilesLeaseNeverRenewed` is the `RenewAgeSecs` sentinel for a lease that has
never been renewed.

At least one export id is required, because version 1 of the gateway's
admin-socket protocol reads one named export at a time and cannot enumerate
what a daemon holds. The client refuses an empty list, and one over the cap,
before sending the request.

## Tests

- `go test -race ./...`
- `go test ./...`
- `GOOS=linux GOARCH=386 go vet ./...`
- `go vet ./...`

`files-exports_test.go` covers the request method, path and query, the
MessagePack decode of a full reply, and the id-count guards on both sides of
the cap. `files-exports_gen_test.go` is the generated round-trip coverage.

## Summary by CodeRabbit

- **New Features**
  - Added an administrative API for querying file export status by export ID.
  - Responses include lease, capacity, node reachability, serving status, totals, and unreachable-node details.
  - Queries support up to 512 export IDs per request with validation for invalid request sizes.

- **Bug Fixes**
  - Improved handling of absent lease renewal timestamps and consistent UTC time values.

- **Documentation**
  - Updated guidance for serialization tooling and wire-format conventions.


files-exports_gen.go
```diff
@@ -0,0 +1,1771 @@
...
+			var zb0002 uint32
+			zb0002, err = dc.ReadArrayHeader()
+			if err != nil {
+				err = msgp.WrapError(err, "Results")
+				return
+			}
+			if cap(z.Results) >= int(zb0002) {
+				z.Results = (z.Results)[:zb0002]
+			} else {
+				z.Results = make([]FilesNodeStatus, zb0002)
```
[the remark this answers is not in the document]
HJLebbink: Not fixing this here. The reading of the generated code is right, but the fix does not belong in this pull request.

**The finding is accurate.** `dc.ReadArrayHeader` returns an unbounded `uint32`, and the generated decoder sizes the slice from that value before it reads a single element. A declared `array32` length of `0xffffffff` makes `make` exceed `maxAlloc` and panic on 64-bit, and on 386 `int(0xffffffff)` is `-1`, so the `cap` test passes and the slice expression panics.

**The shape is not specific to these six sites.** `msgp` emits it for every slice field in the module: 284 `ReadArrayHeader` and `ReadArrayHeaderBytes` call sites across 22 `_gen.go` files. This pull request adds 6 of them, in the same form as the other 278.

**None of the three suggested routes is available in this pull request:**

| Route | Why it does not work |
| --- | --- |
| Bound through generator configuration, then regenerate | `msgp` v1.6.4 has no directive that caps a slice length. `//msgp:` offers `tuple`, `ignore`, `shim`, `replace` and `size`, none of which bound a decoded array header. |
| Hand-write a bounded decoder outside the generated file | It would have to replace the generated `DecodeMsg` and `UnmarshalMsg` for `FilesExportsQueryResponse`, and Go permits one method of each name. CI then fails: the lint job runs `go generate ./...` and rejects any `_gen.go` diff. |
| Limit the response body with `io.LimitReader` | It does not defend against this. The allocation happens on the array header, before any element byte is read, so a 5-byte header still triggers it. |

**Where the fix belongs.** Bounding this means changing the generator, or adding a guard that `msgp` emits for every type. That is one change covering all 284 sites, and a separate pull request against this repository or against `tinylib/msgp`.

**On the threat model.** The bytes decoded here come from a MinIO server the caller configured and authenticated against with a signed request. A server able to send a hostile array header already holds the caller's credentials and data.

HJLebbink:
Status of the two open items.

**1. The failing lint job is fixed.** The `go:generate` line named `msgp` directly, while every other file in the repository calls it through the module tool directory:

```
-//go:generate msgp -d clearomitted -d "timezone utc" -file $GOFILE
+//go:generate go tool msgp -d clearomitted -d "timezone utc" -file $GOFILE
```

`msgp` is a tool dependency in `go.mod`, not a binary on `PATH`, so the bare form failed the `go generate ./...` step with `exec: "msgp": executable file not found in $PATH`. The generated files are unchanged: `go generate ./...` now completes and leaves no `_gen.go` diff. `go build ./...`, `go vet ./...` and `go test -race ./...` all pass.

**2. The docstring coverage warning is a false signal.** The check counted 39 functions across the four files. One of them is hand-written: `AdminClient.FilesExportsQuery`, which carries a doc comment, as does every exported type and field in `files-exports.go`. The remaining functions are the `msgp` methods in `files-exports_gen.go` and the round-trip tests in `files-exports_gen_test.go`, both produced by `go tool msgp`. Documenting them means editing generated files, which CI rejects.

**3. The array-bound finding** is answered in its own thread. Short version: the reading is correct, the shape is repo-wide across 284 generated call sites, and it cannot be fixed for one type without hand-editing generated code.

harshavardhana:
You can add the tool installation for msgp like how we do in `aistor` @HJLebbink that would help resolve the conundrum.

.github/workflows/lint.yml
```diff
@@ -40,6 +40,9 @@ jobs:
           echo "Running go vet..."
           go vet -vettool=$(which staticcheck) ./...
 
+      - name: Install code generation tools
+        run: go install tool
+
```
klauspost: Actually, we should switch to `//go:generate go tool xyz`, so we don't need to install anything - and it will always use the correct version.

You can see in aistor issue 6154. We should probably do that as a separate PR. So I suggest dropping this here.

files-exports.go
```diff
@@ -0,0 +1,241 @@
...
+
+// MaxFilesExportIDsPerQuery is the number of export ids one FilesExportsQuery
+// may name. The server applies the same bound and counts every id named, so
+// repeating an id does not buy a longer list.
+const MaxFilesExportIDsPerQuery = 512
+
+// FilesLeaseNeverRenewed is the FilesExportStatus.RenewAgeSecs value for a
+// lease that has never been renewed. It is the maximum uint64, so comparing the
+// age against TTLSecs treats it as the oldest possible renewal.
+const FilesLeaseNeverRenewed = ^uint64(0)
```
klauspost: I don't like "magic" values - and not in API's in particular. The field should be self-describing.

files-exports.go
```diff
@@ -0,0 +1,241 @@
...
+	// Epoch is the ownership epoch this daemon is serving under.
+	Epoch uint64 `json:"epoch" msg:"e"`
+
+	// OwnerID is an opaque per-process owner identity, for diagnostics only.
+	OwnerID string `json:"ownerID" msg:"o"`
+
+	// RenewAgeSecs is the number of seconds since the last successful lease
+	// renewal, or FilesLeaseNeverRenewed when the lease has never been renewed.
+	// Compare it against TTLSecs to judge whether ownership is at risk.
+	RenewAgeSecs uint64 `json:"renewAgeSecs" msg:"ra"`
```
klauspost: As pointed out above I am not a fan of "magic" values. Using a pointer makes it clear that it was newer renewed.
```suggestion
	RenewAgeSecs *uint64 `json:"renewAgeSecs" msg:"ra,omitempty"`
```

From an API standpoint, I'd rather have an absolute time than a relative.

```suggestion
	LastRenew *time.Time `json:"lastRenew" msg:"lr,omitempty"`
```

files-exports.go
```diff
@@ -0,0 +1,241 @@
...
+// FilesExportStatus is one export's lease and capacity document, as the
+// gateway daemon holding it reported.
+type FilesExportStatus struct {
+	// ExportID is the Ganesha Export_Id.
+	ExportID uint64 `json:"exportID" msg:"id"`
+
+	// Leasing reports whether ownership leasing is configured for this export.
+	// When it is false the other lease fields are zero and only the capacity
+	// fields carry meaning.
+	Leasing bool `json:"leasing" msg:"l"`
```
klauspost: Bools are clean to omit when false...
```suggestion
	Leasing bool `json:"leasing" msg:"l,omitempty"`
```

Since you have the `clearomitted` directive on the generator, all omitted fields will be set to their zero value when reading the object.

Check the rest. I see it is used inconsistently.

files-exports.go
```diff
@@ -0,0 +1,241 @@
...
+	// means unlimited.
+	UsedBytes  uint64 `json:"usedBytes" msg:"ub"`
+	LimitBytes uint64 `json:"limitBytes" msg:"lb"`
+
+	// UsedInodes and LimitInodes are the same, for inodes.
+	UsedInodes  uint64 `json:"usedInodes" msg:"ui"`
+	LimitInodes uint64 `json:"limitInodes" msg:"li"`
+
+	// TS is the Unix time in seconds at which the document was taken.
+	TS int64 `json:"ts" msg:"ts"`
```
klauspost:
```suggestion
	// TS is the Unix time in seconds at which the document was taken.
	TS time.Time `json:"ts" msg:"ts"`
```

Much easier for callers to handle.

HJLebbink: Fixed all the MSGP issues; I also added a brief note to CLAUDE.md to use these conventions in the future.

klauspost (approved):

PR #664, opened 2026-09-23
feat: add TablesReplicationLagVersions admin client method

0xMALVEE:
## Summary
- Adds `AdminClient.TablesReplicationLagVersions(ctx, keys []string) (map[string]int, error)`, a thin wrapper around the existing `POST /tables/replication-lag/versions` admin API, matching the pattern already used by `TablesReplicationStatus` in the same file.
- AIStor's tables replication lag dashboard has a replica site call this endpoint on the primary peer to resolve its own MetadataVersion for a set of tracker keys. Today that caller hand-rolls the HTTP request (marshal body, execute, bound-read, unmarshal); this gives it (and any other caller) a single typed call instead.
- Used by [miniohq/aistor#7408](https://github.com/miniohq/aistor/pull/7408), which replaces its hand-rolled peer call with this method.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofumpt -l .` (clean)
- [x] `golangci-lint run --timeout 5m --config ./.golangci.yml ./...` (0 issues)

## Summary by CodeRabbit

* **New Features**
  * Added a way to retrieve replication-lag versions for a list of tracker keys.


shtripat (approved):
lgtm

